PR #190, opened 2023-03-13
[NRLF-355] integration tests

jaklinger:


api/producer/firehose/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@pytest.fixture(scope="session")
+def session():
+    return new_aws_session()
+
+
+@pytest.fixture(scope="session")
+def stream_arn():
+    firehose_metadata = get_terraform_json()["firehose"]["value"]["producer"][
```
Rohoolio, line 25: Im assuming there is no way to reduce the repetition of this line through the next 4 functions? Minor thing though as I cant think of something that wouldnt make it more complicated haha

api/producer/firehose/tests/e2e_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    records = [FirehoseSubmissionRecord(Data=cloudwatch_data.encode())]
+    response = _submit_records(firehose_client=firehose_client, records=records)
+    assert response.failed_put_count == 0, "This is probably a transient error"
+    return datetime.utcnow()
+
+
+@cache
+def _parse_file(s3_client, bucket_name, object_key) -> list[dict]:
+    response = s3_client.get_object(Bucket=bucket_name, Key=object_key)
+    print("Retrieving data from ", bucket_name, object_key)
```
Rohoolio, line 81: Do these work with your future piece of fixing print statements?

api/producer/firehose/tests/conftest.py
jaklinger: i could write another function from which these fixtures draw from, but i think it would have little value in terms of readability. These wouldn't be necessary if I wasn't using fixtures for these constants, but I'm only using them since it makes the pytest itself a whole lot more readable with a trade off that there are magic fixtures sitting in the conftest. Bit of a tricky one

api/producer/firehose/tests/e2e_utils.py
jaklinger: print statements are ok in non prod code - in pytests anyway they all get captured unless there's a failure, and in this particular case it helps to debug missing firehose files

api/producer/firehose/tests/test_producer_firehose_index.py
```diff
@@ -0,0 +1,102 @@
...
+
+@pytest.mark.slow
+@pytest.mark.integration
+@pytest.mark.firehose
+def test_firehose_output(
+    session,
+    bucket_name,
+    prefix_template,
+    error_prefix_template,
+    global_event_handler,
```
Rohoolio, line 38: I think im being dense but how does this input work from the conftest file? is it because the value was set to pytest session?

api/producer/firehose/tests/test_producer_firehose_index.py
```diff
@@ -0,0 +1,102 @@
...
+    _submit_bad_cloudwatch_data,
+    _submit_very_bad_cloudwatch_data,
+):
+    assert (
+        len(global_event_handler) == 3
+    )  # Sanity check: cloudwatch events have been submitted
+
+    # Some setup
+    start_time = datetime.utcnow()
+    possible_prefixes = set(
```
Rohoolio, line 49: Cant work out in my head what these possible prefixes would look like

Rohoolio: ok could have just read a few lines lower lol

layer/nrlf/nrlf/core/firehose/model.py
```diff
@@ -60,7 +60,7 @@ class CloudwatchLogsData(BaseModel):
 
     # Fields we need
     record_id: str = Field(exclude=True)  # For convenience, not part of the AWS model
-    log_events: list[dict] = Field(alias="logEvents")
+    log_events: conlist(dict, min_items=1) = Field(alias="logEvents")
```
Rohoolio, line 63: Didnt know you could do this, i like it

api/producer/firehose/tests/test_producer_firehose_index.py
```diff
@@ -0,0 +1,102 @@
...
+    # NB the following loops for 600 secs then fails unless
+    # the final break statement is executed
+    for prefix, file_data in retrieve_firehose_output(
+        s3_client=session.client("s3"),
+        bucket_name=bucket_name,
+        start_time=start_time,
+        possible_prefixes=possible_prefixes,
+    ):
+        if prefix.startswith("processed"):
+            if (not verify_good_cloudwatch_data) and all(
```
Rohoolio, line 69: For the sake of readability a bit of function refactoring might be good here as its quite hard to discern whats going on with all the different ifs nots and alls

api/producer/firehose/tests/test_producer_firehose_index.py
jaklinger: when pytest runs any test, it looks for the arguments from matching fields in `parametrize` and from matching fields from any available `fixture`. Fixtures are in scope if they're in the same module, or a `conftest.py` in the same directory as the test, or the same directory from where `pytest` is called

api/producer/firehose/tests/test_producer_firehose_index.py
jaklinger: yep fair

Rohoolio (approved):

jaklinger:
Superseded by `feature/NRLF-355-firehose_setup` for the release branch

PR #207, opened 2023-03-22
[NRLF-376] Final page now always returns no token

jaklinger:


layer/nrlf/nrlf/core/repository.py
```diff
@@ -75,22 +75,22 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def _valid_results(
-    item_type: type[DynamoDbModel], results: list[dict]
+def _valid_items(
```
jaklinger, line 78: Only reason I changed this was to be consistent with language i was using when calling the function

Rohoolio (approved):

PR #213, opened 2023-03-28
[NRLF-371] Removed validated of nhs_number in _NHSNumberMixin

brianbayes:


layer/nrlf/nrlf/core/model.py
```diff
@@ -244,7 +244,6 @@ def nhs_number(self) -> Union[str, None]:
             return None
         try:
             nhs_number = self.subject_identifier.__root__.split("|", 1)[1]
-            validate_nhs_number(nhs_number)
```
Rohoolio, line 247: I thought we wanted to be more restrictive on the NHS number? Unless I remember the previous conversations wrong we would always be using the NHS number as the identifier so the flexibility of the create endpoint was the issue?

PR #229, opened 2023-04-04
[NRLF-421] test-smoke-script 

Rohoolio:
refactored smoke test and command

### DIFF
--- a/.github/workflows/persistent-environment.yml
+++ b/.github/workflows/persistent-environment.yml
@@ -27,8 +27,8 @@ permissions:
   actions: write
 
 env:
-  ENVIRONMENT: ${inputs.environment}
-  WORKSPACE: ${ENVIRONMENT}${{ github.event.inputs.sandbox_deploy == 'true' && '-sandbox' || '' }}
+  ENVIRONMENT: ${{ inputs.environment }}
+  WORKSPACE: ${{ inputs.environment }}${{ github.event.inputs.sandbox_deploy == 'true' && '-sandbox' || '' }}
   TF_WORKFLOW_ID: ${{ inputs.environment }}
   CACHE_NAME: ${{ inputs.environment }}-cache
   ARTIFACT_NAME: ${{ inputs.environment }}-artifact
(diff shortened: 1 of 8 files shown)


api/tests/test_smoke.py
```diff
@@ -157,13 +176,11 @@ def _prepare_base_request(env: str, actor: str) -> tuple[str, dict]:
...
     session = aws_session_assume_terraform_role(account_id)
 
-    apigee_base_url = aws_read_ssm_apigee_proxy(session, env)
-    if apigee_base_url.strip() == "":
-        pytest.skip("Smoke Test Disabled: No APIGEE Proxy defined")
-
     oauth_token = get_oauth_token(session, account, env)
 
-    base_url = f"https://{apigee_base_url}/nrl-{actor}-api"
+    # base_url = f"https://{apigee_env}.{APIGEE_BASE_URL}/record-locator/{actor}"
```
brianbayes, line 181: Probably remove commented-out line :)

brianbayes (approved):

.github/workflows/prod-environment.yml
```diff
@@ -206,7 +206,7 @@ jobs:
         run: |
           source .venv/bin/activate
           source nrlf.sh
-          nrlf test smoke
+          nrlf test smoke producer ${{ inputs.environment }}
```
jaklinger, line 209: we don't want ${{ inputs.environment }} here do we?

.github/workflows/persistent-environment.yml
```diff
@@ -217,7 +217,7 @@ jobs:
         run: |
           source .venv/bin/activate
           source nrlf.sh
-          nrlf test smoke
+          nrlf test smoke producer ${{ inputs.environment }}${{ github.event.inputs.sandbox_deploy == 'true' && '-sandbox' || '' }}
```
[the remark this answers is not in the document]
Rohoolio: In the original smoke tests it was hardcoded against producer so ive kept that the same for now as handling producer vs consumer i think would make a bit more work than expected

.github/workflows/prod-environment.yml
Rohoolio: yeah you still need to provide the environment for the command so it can get the right oauth details from secrets manager

jaklinger (approved):

PR #239, opened 2023-04-12
[NRLF-438] duplicate-json-keys 

Rohoolio:
Added new json loads function to handle duplicate keys

### DIFF
--- a/layer/nrlf/nrlf/core/validators.py
+++ b/layer/nrlf/nrlf/core/validators.py
@@ -6,8 +6,8 @@ from nrlf.core.constants import ID_SEPARATOR, VALID_SOURCES
 from nrlf.core.errors import (
     AuthenticationError,
     DocumentReferenceValidationError,
-    FhirValidationError,
     DuplicateKeyError,
+    FhirValidationError,
     InvalidTupleError,
 )
 from nrlf.legacy.constants import NHS_NUMBER_SYSTEM_URL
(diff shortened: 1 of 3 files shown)


layer/nrlf/nrlf/core/tests/test_event_parsing.py
```diff
@@ -19,17 +19,16 @@ def test_parsing_body_from_event(event, expected_body):
...
 
 @pytest.mark.parametrize(
-    "event",
+    ["event", "error_message"],
     [
-        make_aws_event(body=None),
-        make_aws_event(body=12345),
-        make_aws_event(body="12345"),
-        make_aws_event(body=True),
+        [make_aws_event(body="12345"), "Body is not expected json type"],
```
jaklinger, line 24: minor comment is not that we're not testing the other error message here, could add

```suggestion
        [make_aws_event(body="{not json]"), "Body is not valid json"],
        [make_aws_event(body="12345"), "Body is not expected json type"],            
```

layer/nrlf/nrlf/core/validators.py
```diff
@@ -111,3 +112,17 @@ def validate_type_system(type: RequestQueryType, pointer_types: list[str]):
...
         )
+
+
+def dict_raise_on_duplicates(list_of_pairs):
+    checked_pairs = {}
+    for k, v in list_of_pairs:
+        if k in checked_pairs:
+            raise DuplicateKeyError("Duplicate key: %r" % (k,))
+        else:
+            checked_pairs[k] = v
```
jaklinger, line 123:
```suggestion
        checked_pairs[k] = v
```

jaklinger (approved):

PR #266, opened 2023-05-03
[NRLF-395] Logic for one-directional-sync header and the created_on date

brianbayes:


feature_tests/features/producer/producer-createDocumentPointer-success.feature
```diff
@@ -160,3 +161,73 @@ Feature: Producer Create Success scenarios
       | updated_on  | NULL                                    |
       | document    | <document>                              |
       | created_on  | <timestamp>                             |
+
+  Scenario: Successfully create a Document Pointer for one directional data sync when date is provided
+    Given Producer "BaRS (EMIS)" (Organisation ID "V4T0L.YGMMC") is requesting to create Document Pointers
```
Rohoolio, line 166: Might be worth changing the names of the provider to something like Data Sync to avoid any confusion in the future seeing as this is a one off event

feature_tests/features/producer/producer-createDocumentPointer-success.feature
```diff
@@ -34,7 +34,8 @@ Feature: Producer Create Success scenarios
             }
           }
         ],
-        "status": "current"
+        "status": "current",
+        "date": "2023-05-02T12:00:00.000Z"
```
Rohoolio, line 38: What happens if this isnt in the template?

feature_tests/features/producer/producer-createDocumentPointer-success.feature
```diff
@@ -160,3 +161,73 @@ Feature: Producer Create Success scenarios
       | updated_on  | NULL                                    |
       | document    | <document>                              |
       | created_on  | <timestamp>                             |
+
+  Scenario: Successfully create a Document Pointer for one directional data sync when date is provided
```
deepaattiguppe, line 165: I am not able to see any differences between the below scenarios apart from the scenario titles? Am i missing something here?     

- Successfully create a Document Pointer for one directional data sync when no date is provided
- Successfully create a Document Pointer for one directional data sync when date is provided

feature_tests/features/producer/producer-createDocumentPointer-success.feature
```diff
@@ -160,3 +161,73 @@ Feature: Producer Create Success scenarios
...
+    Given Producer "BaRS (EMIS)" (Organisation ID "V4T0L.YGMMC") is requesting to create Document Pointers
+    And Producer "BaRS (EMIS)" is registered in the system for application "DataShare" (ID "z00z-y11y-x22x") with pointer types
+      | system                 | value           |
+      | http://snomed.info/sct | 861421000000109 |
+    And Producer "BaRS (EMIS)" is one directional data sync
```
deepaattiguppe, line 170: Can we have a different producer name apart from BaRS(EMIS) ? maybe OneDirectionalDataSync as it will improve readability of the tests

api/producer/createDocumentReference/src/v1/handler.py
```diff
@@ -52,6 +53,17 @@ def _invalid_type(
     return source_document_pointer.type != target_document_pointer.type
 
 
+def _override_created_on(
+    data: PipelineData, document_pointer: DocumentPointer
+) -> DocumentPointer:
+    fhir_model: StrictDocumentReference = create_fhir_model_from_fhir_json(
+        fhir_json=data["body"]
+    )
+    if fhir_model.date is not None:
```
Rohoolio, line 62: Minor point but if we dont receive a date (eg this value is none) do we actually care about that or can we do as we do here and just carry on?

brianbayes: It's potentially a redundant check tbh as setting the `fhir_model` will blow up if you don't provide it a date as part of the `create_fhir_model_from_fhir_json` method.

PR #273, opened 2023-05-11
[NRLF-512] sync-supersede-date

Rohoolio:
- if audit date permission present both the superseded document or the created document will have a create_on based on the date provided in the payload

### DIFF
--- a/feature_tests/steps/_1_setup.py
+++ b/feature_tests/steps/_1_setup.py
@@ -93,7 +93,7 @@ def has_permissions(
     test_config: TestConfig = context.test_config
     existing_headers = json_loads(test_config.request.headers[CONNECTION_METADATA])
 
-    permissions = [f'{row["permission"]}' for row in context.table]
+    permissions = [row["permission"] for row in context.table]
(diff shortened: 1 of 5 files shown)


feature_tests/steps/_1_setup.py
```diff
@@ -84,6 +84,21 @@ def has_permissions(
...
+@given('{actor_type} "{actor}" has the permissions')
+def has_permissions(
+    context: Context,
+    actor_type: str,
+    actor: str,
+):
+    test_config: TestConfig = context.test_config
+    existing_headers = json_loads(test_config.request.headers[CONNECTION_METADATA])
+
+    permissions = [f'{row["permission"]}' for row in context.table]
```
jaklinger, line 96:
```suggestion
    permissions = [row["permission"] for row in context.table]
```

unless this is to convert to string, in which case the explicit method is more direct:

```suggestion
    permissions = [str(row["permission"]) for row in context.table]
```

jaklinger (approved):

PR #290, opened 2023-05-24
[NRLF-535] New Count Endpoint

nomad3k:
Added new `GET /DocumentReference/_count` endpoint

* Updates to Swagger
* New countDocumentReference lambda
* New success and failure feature tests


api/producer/countDocumentReference/src/v1/handler.py
```diff
@@ -0,0 +1,71 @@
...
+from nrlf.core.errors import assert_no_extra_params
+from nrlf.core.model import PaginatedResponse, ProducerRequestParams, key
+from nrlf.core.repository import Repository, type_filter
+from nrlf.core.transform import create_bundle_count
+from nrlf.core.validators import validate_type_system
+from nrlf.producer.fhir.r4.model import NextPageToken, RequestQuerySubject
+
+
+class LogReference(Enum):
+    COUNT_001 = "Searching for document references"
```
jaklinger, line 20: shouldn't this just be `COUNT` rather than `FOO_001`?

api/producer/countDocumentReference/src/v1/handler.py
```diff
@@ -0,0 +1,71 @@
...
+        pointer_types=data["pointer_types"],
+    )
+
+    count = 0
+    next_page_token: NextPageToken = None
+    while True:
+        response: PaginatedResponse = repository.query_gsi_2(
+            pk=key(DbPrefix.Organization, *ods_code_parts),
+            type=pointer_types,
+            nhs_number=nhs_number,
```
Rohoolio, line 56: Will be worth putting a custom limit here to avoid this process circling around the default page limit of 20

limit=1000 

I think 1000 is the currently highest amount but it could save us going out and in since we just want the count

api/producer/countDocumentReference/src/v1/handler.py
```diff
@@ -0,0 +1,71 @@
...
+from nrlf.core.validators import validate_type_system
+from nrlf.producer.fhir.r4.model import NextPageToken, RequestQuerySubject
+
+
+class LogReference(Enum):
+    COUNT_001 = "Searching for document references"
+
+
+@log_action(log_reference=LogReference.COUNT_001)
+def count_document_references(
```
jaklinger, line 24: might be worth adding the countDocumentReference endpoints to this ticket [NRLF-460](https://nhsd-jira.digital.nhs.uk/browse/NRLF-460) to be refactored at a later stage, given the overlap

api/producer/countDocumentReference/src/v1/handler.py
Rohoolio: Its more to do with the payload size actually - so you can make that number as big as you want and it will stop at 1mb size

api/consumer/record-locator/consumer.yaml
```diff
@@ -215,6 +215,7 @@ paths:
         - $ref: "#/components/parameters/type"
         - $ref: "#/components/parameters/nextPageToken"
         - $ref: "#/components/parameters/odsCode"
+        - $ref: "#/components/parameters/odsCodeExtension"
```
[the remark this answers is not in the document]
nomad3k: Because the `main` branch was out of sync, and running `nrlf swagger generate` corrected this.

api/producer/countDocumentReference/src/v1/handler.py
nomad3k: Yes

api/producer/countDocumentReference/src/v1/handler.py
nomad3k: We can do that when we elaborate NRLF-460.

api/producer/countDocumentReference/src/v1/handler.py
nomad3k: So I upped it to 1000, but this made the testing requirement become at least 2001 records, which made the feature tests significantly slower.  Therefore I opted for a compromise and upped it to 100, which was larger but still performant during CI/CD.

api/producer/countDocumentReference/src/v1/handler.py
```diff
@@ -0,0 +1,72 @@
...
+
+    count = 0
+    next_page_token: NextPageToken = None
+    while True:
+        response: PaginatedResponse = repository.query_gsi_2(
+            pk=key(DbPrefix.Organization, *ods_code_parts),
+            type=pointer_types,
+            nhs_number=nhs_number,
+            exclusive_start_key=next_page_token,
+            limit=100,
```
Rohoolio, line 58: Could we give this a name rather than it being a magic number?

nomad3k: Agreed.  Done.

Rohoolio (dismissed):

PR #291, opened 2023-05-25
[NRLF-539] asid-author

Rohoolio:
Updated the feature tests with the new converter changes to asid in the author


jaklinger (approved):
looks great

PR #320, opened 2023-06-15
[NRLF-506a] db query

jaklinger:


terraform/account-wide-infrastructure/mgmt/rds-cluster-vpc.tf
```diff
@@ -0,0 +1,88 @@
+data "aws_availability_zones" "available" {}
```
Rohoolio, line 1: Im fairly sure we wont need any of this in mgmt

PR #339, opened 2023-06-21
[NRLF-533] Supersede error message replaced and http status code

brianbayes:


feature_tests/features/producer/producer-supersedeDocumentPointer-failure.feature
```diff
@@ -1,4 +1,4 @@
-Feature: Producer Supersede Failure scenarios
+Feature: Producer Supersede Edge Case scenarios
```
jaklinger, line 1: Don't we still need all of the failure tests? Also not sure that the new error message is being tested here is it?

layer/nrlf/nrlf/core/repository.py
```diff
@@ -494,8 +493,8 @@ def update(self, item: PydanticModel) -> DynamoDbResponse:
         return self.dynamodb.put_item(**args)
 
     @handle_dynamodb_errors(
-        conditional_check_error_message="Supersede ID mismatch",
-        error_type=SupersedeError,
+        conditional_check_error_message="Duplicate item",
+        error_type=DuplicateError,
```
jaklinger, line 497: Thing is that this now doesn't provide the Supersede ID Mismatch error which i think we still want for other scenarios. The problem with `handle_dynamodb_errors` is that it's a catch all - but really that a reflection of the fact that dynamo doesn't tell you what went wrong.

I think the only way to fully resolve the ticket is to make `handle_dynamodb_errors` a bit smarter somehow. I don't think it's necessarily an easy job, or even feasible to do nicely. I'd argue that the ticket requires a spike in general to understand how to return nuanced errors from dynamo. I'd say that it's almost probably not worth doing...

feature_tests/features/producer/producer-supersedeDocumentPointer-failure.feature
brianbayes: No idea how this ended up here as 'Edge Cases'

PR #344, opened 2023-06-22
[feature/NRLF-XXXX-auth_from_s3] Auth from S3

jaklinger:


feature_tests/common/constants.py
```diff
@@ -59,6 +60,9 @@
     DocumentPointer: DOCUMENT_POINTER_TABLE_DEFINITION,
 }
 
+PERMISSIONS_BUCKET = "permissions_bucket"  # This should match the value in TF outputs
```
Rohoolio, line 63: Is this value right or does it get changed later? its name is now 

"${local.prefix}--authorization-store"

layer/lambda_utils/lambda_utils/header_config.py
```diff
@@ -41,10 +41,13 @@ class ClientRpDetailsHeader(AbstractHeader):
 
 
 class ConnectionMetadata(AbstractHeader):
-    pointer_types: list[str] = Field(alias="nrl.pointer-types")
+    pointer_types: list[str] = Field(alias="nrl.pointer-types", default_factory=list)
     ods_code: str = Field(alias="nrl.ods-code")
     ods_code_extension: str = Field(alias="nrl.ods-code-extension", default=None)
-    nrl_permissions: list[str] = Field(alias="nrl.permissions", default=[])
+    nrl_permissions: list[str] = Field(alias="nrl.permissions", default_factory=list)
+    enable_permissions_lookup: bool = Field(
```
Rohoolio, line 48: Im not sure the custom attribute can be a bool - will it be able to parse this?

layer/nrlf/nrlf/core/authoriser.py
```diff
@@ -42,9 +54,12 @@ class Config(BaseModel):
     ENVIRONMENT: str
     SPLUNK_INDEX: str
     SOURCE: str
+    PERMISSIONS_LOOKUP_BUCKET: str
```
Rohoolio, line 57: I think we want to change this to be Authorization store to match the Terraform now

layer/nrlf/nrlf/core/authoriser.py
```diff
@@ -58,6 +73,8 @@ def build_persistent_dependencies(config: Config) -> dict[str, any]:
         "environment": config.ENVIRONMENT,
         "splunk_index": config.SPLUNK_INDEX,
         "source": config.SOURCE,
+        "permissions_lookup_bucket": config.PERMISSIONS_LOOKUP_BUCKET,
```
Rohoolio, line 76: Same as above

layer/nrlf/nrlf/core/authoriser.py
Rohoolio: Which I should have done when I committed my other PR haha

layer/nrlf/nrlf/core/tests/test_authoriser_utils.py
```diff
@@ -79,6 +142,8 @@ def test_execute_steps_yields_deny_on_internal_server_error(
             ENVIRONMENT="an-environment",
             SPLUNK_INDEX="an-index",
             SOURCE="the-lambda-name",
+            S3_CLIENT="the-s3-client",
+            PERMISSIONS_LOOKUP_BUCKET="a-bucket",
```
Rohoolio, line 146: Same here

terraform/infrastructure/lambda.tf
```diff
@@ -266,11 +266,13 @@ module "producer__authoriser_lambda" {
   layers      = [module.lambda-utils.layer_arn, module.nrlf.layer_arn, module.third_party.layer_arn]
   kms_key_id  = module.kms__cloudwatch.kms_arn
   environment_variables = {
-    PREFIX       = "${local.prefix}--"
-    ENVIRONMENT  = local.environment
-    SPLUNK_INDEX = module.firehose__processor.splunk.index
+    PREFIX                    = "${local.prefix}--"
+    ENVIRONMENT               = local.environment
+    SPLUNK_INDEX              = module.firehose__processor.splunk.index
+    PERMISSIONS_LOOKUP_BUCKET = aws_s3_bucket.authorization-store.id
```
Rohoolio, line 272: This is where we want to change the env var to something more auth store related

terraform/infrastructure/lambda.tf
```diff
@@ -289,9 +291,10 @@ module "consumer__authoriser_lambda" {
...
```
Rohoolio, line 297: Nice catch - i did not remember to put the same env var in the consumer 😓

terraform/infrastructure/output.tf
```diff
@@ -57,3 +57,7 @@ output "firehose" {
 output "mi" {
   value = module.mi
 }
+
+output "permissions_bucket" {
```
Rohoolio, line 61: Again ive called this permissions bucket but we should probably adjust that to be consistent to authorization_store

feature_tests/common/constants.py
jaklinger: oh that comment's old actually, now it's just a dummy name for the local run (i.e. when mocking s3)

jaklinger: so i'll rm the comment

layer/lambda_utils/lambda_utils/header_config.py
jaklinger: Seems to work fine:

image

jaklinger: the value we pass to NRLF can certainly be a bool, more of a question about how we choose to parse this on the apigee side

layer/nrlf/nrlf/core/authoriser.py
jaklinger: ah yep

PR #371, opened 2023-07-03
[NRLF-460] Merge steps for search rebased

axelkrastek1-nhs:


api/consumer/countDocumentReference/src/v1/handler.py
```diff
@@ -37,43 +30,21 @@ def search_document_references(
...
-            exclusive_start_key=next_page_token,
-            limit=COUNT_ITEM_LIMIT,
+        response: PaginatedResponse = get_paginated_document_references(
+            data,
+            request_params,
+            event.queryStringParameters,
+            dependencies,
+            logger,
+            COUNT_ITEM_LIMIT,
+            next_page_token,
```
jaklinger, line 47: for a function with this many params we should be using keyword arguments - can come back to bite us if we need to refactor positional args

api/consumer/searchDocumentReference/src/v1/handler.py
```diff
@@ -32,41 +24,13 @@ def search_document_references(
...
-    custodian = custodian_filter(
-        custodian_identifier=request_params.custodian_identifier
-    )
-
-    validate_type_system(request_params.type, pointer_types=data["pointer_types"])
-
-    pointer_types = type_filter(
-        type_identifier=request_params.type,
-        pointer_types=data["pointer_types"],
+    response = get_paginated_document_references(
```
jaklinger, line 27: as above re: positional args

api/consumer/searchPostDocumentReference/src/v1/handler.py
```diff
@@ -34,37 +25,11 @@ def search_document_references(
...
-    request_params = ConsumerRequestParams(**body)
-    assert_no_extra_params(request_params=request_params, provided_params=body)
-    nhs_number: RequestQuerySubject = request_params.nhs_number
-    pk = key(DbPrefix.Patient, nhs_number)
 
-    custodian = custodian_filter(
-        custodian_identifier=request_params.custodian_identifier
+    response = get_paginated_document_references(
+        data, ConsumerRequestParams(**body), body, dependencies, logger
     )
```
jaklinger, line 32: probably best to have `request_params = ConsumerRequestParams(**body)` on it's own line so that if an error is thrown the message is more informative

jaklinger: also same comment re: positional args

layer/nrlf/nrlf/core/common_search_steps.py
```diff
@@ -0,0 +1,92 @@
...
+
+class LogReference(Enum):
+    COMMONSEARCH001 = "Searching for document references"
+
+
+@log_action(log_reference=LogReference.COMMONSEARCH001)
+def get_paginated_document_references(
+    data: PipelineData,
+    requestParams: Union[ConsumerRequestParams, CountRequestParams],
+    queryStringParams: Dict[str, str],
```
jaklinger, line 38: camel case for params isn't standard - could you switch to snake?

layer/nrlf/nrlf/core/common_search_steps.py
```diff
@@ -0,0 +1,92 @@
...
+    dependencies: FrozenDict[str, Any],
+    logger: Logger,
+    pageLimit: int = PAGE_ITEM_LIMIT,
+    pageToken: Annotated[
+        Optional[NextPageToken], Field(alias="next-page-token")
```
jaklinger, line 43: same re: snake

jaklinger, line 43: `Field(alias="next-page-token")` isn't doing anything here

layer/nrlf/nrlf/core/common_search_steps.py
```diff
@@ -0,0 +1,92 @@
...
+
+log_action = make_common_log_action()
+
+
+class LogReference(Enum):
+    COMMONSEARCH001 = "Searching for document references"
+
+
+@log_action(log_reference=LogReference.COMMONSEARCH001)
+def get_paginated_document_references(
```
jaklinger, line 35: possible to split this function into a few parts? would be better for logging and debugging

layer/nrlf/nrlf/core/common_search_steps.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+
+    if isinstance(requestParams, ConsumerRequestParams):
+        if pageToken is None:
+            next_page_token: NextPageToken = request_params.next_page_token
+        if next_page_token is not None:
+            next_page_token = next_page_token.__root__
+
+    else:
+        next_page_token: NextPageToken = pageToken
```
jaklinger, line 80: better to use dependency injection here rather than branching logic in the common function

jaklinger (approved):

PR #397, opened 2023-07-13
[NRLF-586] Changelog consolidate publish

jameslinnell:


changelog/scripts/changelog.py
```diff
@@ -0,0 +1,48 @@
...
+from os import listdir
+from os.path import getsize, isfile, join
+
+
+def main(argv):
+    if "release/" in argv[0]:
+        file_list = sorted(
+            [f for f in listdir("./changelog") if isfile(join("./changelog", f))]
+        )
+        file_list.reverse()
```
jaklinger, line 11: just an fyi that there is also `sorted(..., reverse=True)`, not that it makes a difference!

jaklinger (approved):

PR #434, opened 2023-08-01
[NRLF-533] supersede when ignore delete fail

nomad3k:


layer/nrlf/nrlf/core/repository.py
```diff
@@ -525,7 +525,15 @@ def _delete(id):
...
-        return self.dynamodb.transact_write_items(TransactItems=transact_items)
+        try:
+            return self.dynamodb.transact_write_items(TransactItems=transact_items)
+        except ClientError as error:
+            if "CancellationReasons" in error.response:
+                reasons = error.response["CancellationReasons"]
+                # Last reason relates to the single PUT and indicates duplicate item
+                if reasons[-1]["Code"] == "ConditionalCheckFailed":
+                    raise DuplicateError("Condition check failed - Duplicate item")
+            raise error
```
jaklinger, line 536: because the `reasons[-1]` and `[put()]` are implicitly coupled this will be prone to a bug in future. I can imagine if, for some reason, someone change L527 to:

```python
transact_items = [_delete(id) for id in delete_pks] + [_put()] + [some_other_transaction()]
```

then `reasons[-1]` wouldn't be related to the put error. This is also generally the problem with `handle_dynamodb_errors`. Since the transactions are annotated with `Put` and `Delete`, I might suggest in this case that we pull the index out of `transact_item` e.g.

```python
def _is_put(transaction: dict):
    action, = transaction.keys()
    return action == "PUT"

put_idx = next(idx for idx, item in enumerate(transact_items) if _is_put(transaction=item))
```

and then 

```python
reasons[put_idx]
```

jaklinger, line 536: to avoid the comment could wrap this up in a context manager function

```python
with handler_put_duplicate_error(put_idx=put_idx):
    return self.dynamodb.transact_write_items(TransactItems=transact_items)
```

feature_tests/features/producer/producer-supersedeDocumentPointer-failure.feature
```diff
@@ -588,3 +624,84 @@ Feature: Producer Supersede Failure scenarios
       | issue_description | Invalid value                           |
       | message           | Condition check failed - Duplicate item |
     And Document Pointer "DS123-1234567890" still exists
+
+  Scenario: Supersede without supersede-ignore-delete-fail already exists with invalid target
```
jaklinger, line 628:
```suggestion
  Scenario: Supersede without 'supersede-ignore-delete-fail' permission a DocumentPointer that exists with invalid target
```

feature_tests/features/producer/producer-supersedeDocumentPointer-failure.feature
```diff
@@ -588,3 +624,84 @@ Feature: Producer Supersede Failure scenarios
...
+    And the response is an OperationOutcome according to the OUTCOME template with the below values
+      | property          | value                                                         |
+      | issue_type        | processing                                                    |
+      | issue_level       | error                                                         |
+      | issue_code        | VALIDATION_ERROR                                              |
+      | issue_description | A parameter or value has resulted in a validation error       |
+      | message           | Validation failure - relatesTo target document does not exist |
+    And Document Pointer "DS123-ALREADY_EXISTS" still exists
+
+  Scenario: Supersede without supersede-ignore-delete-fail already exists with valid target
```
jaklinger, line 664:
```suggestion
  Scenario: Supersede without 'supersede-ignore-delete-fail' permission a DocumentPointer that exists with valid target
```

jaklinger (changes requested):

layer/nrlf/nrlf/core/repository.py
nomad3k: Or we add a comment above `[put()]`

nomad3k: We can also search through the `Item` on the `reasons` to see if it's `id` matches the subject's `id`.  This will be simpler than the index additions above.

jaklinger (approved):

PR #462, opened 2023-08-10
[NRLF-644] errors in swagger consumer

jameslinnell:


nomad3k (changes requested):
The `/_count` endpoint is missing the Errors, because this is the only endpoint that resides in it's own swagger file.

nomad3k (approved):

api/consumer/swagger.yaml
```diff
@@ -606,6 +697,34 @@ paths:
         - $ref: "#/components/parameters/requestId"
         - $ref: "#/components/parameters/correlationId"
       responses:
+        "4xx":
```
axelkrastek1-nhs, line 700: Should this be uppercase as well?

api/consumer/record-locator/consumer.yaml
```diff
@@ -567,6 +658,34 @@ paths:
         - $ref: "#/components/parameters/requestId"
         - $ref: "#/components/parameters/correlationId"
       responses:
+        "4xx":
```
axelkrastek1-nhs, line 661: Should this be uppercase as well?

swagger/consumer-static/count.yaml
```diff
@@ -12,6 +12,34 @@ paths:
         - $ref: "#/components/parameters/requestId"
         - $ref: "#/components/parameters/correlationId"
       responses:
+        "4xx":
```
axelkrastek1-nhs, line 15: Should this be uppercase as well?

PR #463, opened 2023-08-11
[NRLF-561] ASID contract with helper

samgardner1-nhs:
In order to validate that an ASID is provided for all care plan document types 
We must have an ASID specific document contract that enforces that the ASID Is present in the author field

Note: This PR is a Draft because these changes are based on top of [feature/NRLF-634-data_contract_helper](https://github.com/NHSDigital/NRLF/tree/feature/NRLF-634-data_contract_helper), and so this branch contains all features from NRLF-634 as well. In other words, do not merge both 634 and 561, only merge 561


data_contracts/deploy_contracts/tests/test_deploy_contract.py
```diff
@@ -903,3 +903,47 @@ def test_sync_contracts_e2e(temp_dir):
...
+
+    # Deploy the contracts
+    today = date(year=2000, month=1, day=1)
+    result = sync_contracts(repository=repository, today=today)
+    assert result is True
+
+    # Verify the contracts have been deployed
+    synced_db_contracts = _get_contracts_from_db(repository=repository)
+    assert initial_db_contracts != synced_db_contracts  # i.e. an update has taken place
+    assert len(synced_db_contracts) == 8
```
jaklinger, line 929: this will break after every update, we should do:

```suggestion
    assert len(synced_db_contracts) == len(PATHS_TO_CONTRACTS)
```

data_contracts/deploy_contracts/tests/test_deploy_contract.py
```diff
@@ -903,3 +903,47 @@ def test_sync_contracts_e2e(temp_dir):
...
+            continue
+        assert group.value in [
+            "325691000000100",
+            "887701000000100",
+            "861421000000109",
+            "736373009",
+            "736253002",
+            "1382601000000107",
+        ]
+        assert group.name in ["asidcheck-contract"]
```
jaklinger, line 947: imo the rest of the test is kind of redundant as group from db has already been tested, and there nothing here that checks the effect of a given contract on a pointer type. I'd remove these lines and then we can rely on the feature tests to give us coverage of these values. With the previous suggestion it means that this test won't ever need updating as well, which is nice

feature_tests/features/producer/producer-createDocumentPointer-failure.feature
```diff
@@ -661,3 +703,55 @@ Feature: Producer Create Failure Scenarios
       | issue_code        | VALIDATION_ERROR                                                                                                                                         |
       | issue_description | A parameter or value has resulted in a validation error                                                                                                  |
       | message           | ValidationError raised from Data Contract 'Validate Content Url:1' at 'content[0].attachment.url': 'not-a-url' does not match '^https*://(www.)*\\w+.*$' |
+
+  @integration-only
+  Scenario: Validate a Document Pointer of type Mental health crisis plan using the asid data contract with ssp and no asid
```
jaklinger, line 708: for all of the soft-linked files we should do a `Scenario Outline` (as opposed to `Scenario`) so that each feature test is run on each pointer-type. Otherwise if we were to remove/change one of the files we'd have no coverage on the behaviour for each pointer type

jaklinger (approved):

PR #466, opened 2023-08-11
[NRLF-653] MI stream writer

jaklinger:


mi/reporting/report.py
```diff
@@ -67,6 +68,9 @@ def make_reports(session, env: str, workspace: str = None, partition_key=None):
 
 
 def _make_reports(env: str, workspace: str = None, partition_key=None):
+    if partition_key is not None:
+        partition_key = str(hash_str_to_int(key=partition_key))
```
Rohoolio, line 72: What did we need to do this for?

mi/reporting/tests/test_data/seed_database.py
```diff
@@ -17,16 +23,48 @@
...
+                int,  # i.e transforms this int
+                int,  # to this int
+            ],
+        ],
+    ] = defaultdict(lambda: defaultdict(dict))
+    for i, item in enumerate(data[Measure.name()]):
+        # Assumption here is that test_id + {row_number} will not clash
+        # between tests, which is unlikely locally given the terraform teardown
+        # on every build, and is impossible in CI
+        uniquish_test_id = test_id + i
```
Rohoolio, line 53: Just struggling to follow this, I might need some help going through it 😅

mi/sql_query/sql.py
```diff
@@ -1,14 +1,34 @@
...
 
-def execute_sql_query(cursor, sql: Sql) -> list:
-    query = psycopg2_sql.SQL(sql.statement).format(
-        **{k: psycopg2_sql.Identifier(v) for k, v in sql.identifiers.items()}
+    # Need to split statements up since when using "autocommit"
+    # it is not possible to invoke multiple statements in a single call.
+    # In practice this is needed for the database clear-down in the
+    # database_administration.tf
+    statements: list[str] = (
+        filter(
```
Rohoolio, line 22: So in this case we only filter the statements on the if statement being true? Im not used to seeing it above the if itself haha

mi/stream_writer/event_handling.py
```diff
@@ -0,0 +1,87 @@
...
+@catch_error(log_fields=["document_pointer"])
+def insert_mi_record(
+    record: RecordParams,
+    sql: str,
+    cursor: Cursor,
+    dimension_types: tuple[type[Dimension]] = DIMENSION_TYPES,
+    integrity_error_type: type[IntegrityError] = IntegrityError,
+    not_null_violation: str = NOT_NULL_VIOLATION,
+) -> GoodResponse:
+    insert_record = partial(
```
Rohoolio, line 70: What do we gain from using a partial here? Arent we still passing in the same amount of arguments without anything being substituted?

mi/stream_writer/event_handling.py
```diff
@@ -0,0 +1,87 @@
...
+        connection.rollback()
+        raise
+
+
+@catch_error(log_fields=["document_pointer"])
+def insert_mi_record(
+    record: RecordParams,
+    sql: str,
+    cursor: Cursor,
+    dimension_types: tuple[type[Dimension]] = DIMENSION_TYPES,
```
Rohoolio, line 66: Do we want to have dimension types as something that can be passed into this function? I feel like its going to be a static set of options and if we wanted to increase that we could just update the DIMENSION_TYPES config right?

mi/stream_writer/model.py
```diff
@@ -0,0 +1,171 @@
...
+    image_type: DynamodbEventImageType
+    action: Action
+
+    @property
+    def sql(self):
+        with open(f"{PATH_TO_QUERIES}/{self.action}.sql") as f:
+            return f.read()
+
+
+class Dimension:
```
Rohoolio, line 71: I understand (I think) whats happening here, but a little run through to understand the benefits would be much appreciated!

mi/stream_writer/tests/test_helpers.py
```diff
@@ -0,0 +1,218 @@
...
+            cursor=mocked_cursor, statement=None, params=None, identifiers=None
+        )
+    mocked_cursor.connection.rollback.assert_called()
+
+
+@mock.patch("mi.stream_writer.event_handling.psycopg2_sql")
+def test_execute_sql_success(mocked_sql):
+    mocked_cursor = mock.Mock(spec=Cursor)
+    mocked_cursor.connection = mock.Mock(spec=Connection)
+    mock.Mock(spec=Cursor)
```
Rohoolio, line 132: What does this line achieve that line 130 doesnt? Confused as to what this bit adds

mi/stream_writer/tests/test_model.py
```diff
@@ -0,0 +1,50 @@
...
+VALUE = "def"
+
+document_pointer_strategy = {
+    k: builds(v.type_.__fields__["__root__"].type_)
+    for k, v in DocumentPointer.__fields__.items()
+}
+
+
+@given(
+    **{
```
Rohoolio, line 22: What do these ** achieve?

.tool-versions
```diff
@@ -1,6 +1,6 @@
 awscli latest
 poetry 1.5.1
-jq latest
+jq 1.6
```
jameslinnell, line 3: We might want to leave this as latest. I know you were having issues, but we probably don't want to restrict it.

mi/stream_writer/tests/test_helpers.py
Rohoolio: success

terraform/infrastructure/modules/postgres_cluster/lambda.tf
```diff
@@ -21,3 +21,73 @@ module "rds-cluster-sql-query-lambda" {
...
+  environment_variables = {
+    POSTGRES_DATABASE_NAME = var.environment
+    RDS_CLUSTER_HOST       = data.aws_rds_cluster.rds-cluster.reader_endpoint
+    RDS_CLUSTER_PORT       = data.aws_rds_cluster.rds-cluster.port
+    POSTGRES_USERNAME      = "${var.environment}-write"
+    POSTGRES_PASSWORD      = aws_secretsmanager_secret.write_password.arn
+  }
+  layers = concat(var.layers, [module.psycopg2.layer_arn, module.nrlf.layer_arn, module.lambda_utils.layer_arn])
+  additional_policies = [
+    "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole",
```
jameslinnell, line 41: Might be a bit nicer as a var but it's not a big issue.

Rohoolio (approved):

.tool-versions
jaklinger: reverted

PR #480, opened 2023-08-18
Feature/nrlf 563 sort immutable fields

andrewrae5-nhs:
added sorting to immutable fields, as per [NRLF-563](https://nhsd-jira.digital.nhs.uk/browse/NRLF-563)

### DIFF
--- a/api/producer/updateDocumentReference/src/v1/handler.py
+++ b/api/producer/updateDocumentReference/src/v1/handler.py
@@ -77,13 +77,32 @@ def document_pointer_exists(
     )
 
 
+def _sort_key(key):
+    """
+    Sorts lists and dictionaries recursively to enable comparisons of sorted objects.
+    Dictionaries are sorted by key, and otherwise lists are sorted by value. Lists of
+    dictionaries are sorted by the dictionary key. Dictionaries are transformed into
+    lists of tuples, again noting that the output of this function is intended for
+    deterministic sorted comparisons.
+    """
+    if type(key) is list:
+        return sorted((_sort_key(k) for k in key), key=_sort_key)
+    elif type(key) is dict:
+        return [(k, _sort_key(key[k])) for k in sorted(key.keys())]
+    return key
+
+
+def _keys_are_not_equal(a, b):
+    return _sort_key(a) != _sort_key(b)
+
+
 def _validate_immutable_fields(
     a: dict, b: dict, immutable_fields: set = IMMUTABLE_FIELDS
 ):
     immutable_keys_in_a = immutable_fields.intersection(a.keys())
     immutable_keys_in_b = immutable_fields.intersection(b.keys())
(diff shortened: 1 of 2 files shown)


api/producer/updateDocumentReference/src/v1/handler.py
```diff
@@ -77,13 +77,36 @@ def document_pointer_exists(
...
+        key_values = list(key.items())[0][1]
+
+        if type(key_values) == dict:
+            return dict(sorted(key_values.items()))
+        elif type(key_values) == list:
+            keys_list = []
+            for sub_keys in key_values:
+                keys_list.extend(list(sub_keys.keys()))
+
+            return sorted(key_values, key=lambda item: sorted(keys_list)[0])
```
jaklinger, line 94: would the following not be sufficient (would sort any dictionary by keys then sort all values):

```suggestion
        return {k: _sort_key(key[k]) for k in sorted(key.keys())}
```

api/producer/updateDocumentReference/src/v1/handler.py
```diff
@@ -77,13 +77,36 @@ def document_pointer_exists(
     )
 
 
+def _sort_key(key):
+
+    if type(key) == list:
+        return sorted(key)
```
jaklinger, line 83: does not work for list of dict, or a list of list, etc, i'd suggest:

```suggestion
        return sorted(key, key=_sort_key)
```

api/producer/updateDocumentReference/src/v1/handler.py
jaklinger, line 1: in other words, suggest to flatten function to:

```python
def _sort_key(key):
    if type(key) is list:
        return sorted(key, key=_sort_key)
    elif type(key) is dict:
        return [(k, _sort_key(key[k])) for k in sorted(key.keys())]
    return key
```

api/producer/updateDocumentReference/tests/unit/test_update_document_reference_v1_steps.py
```diff
@@ -81,6 +82,134 @@ def test_compare_immutable_fields_success(mock__make_timestamp):
...
+            "custodian": {"S": "Y05868"},
+            "source": {"S": "NRLF"},
+            "type": {"S": "http://snomed.info/sct|736253002"},
+            "updated_on": {"S": "2022-10-25T15:47:49.732Z"},
+            "version": {"N": "1"},
+            "author": [
+                {"id": "a", "reference": "b"},
+                {"id": "b", "reference": "a"},
+                {"reference": "c", "display": "d"},
+            ],
```
jaklinger, line 167: to fix the broken test:

```suggestion
```

andrewrae5-nhs: whooops! strange, worked before :/

jaklinger (approved):

PR #494, opened 2023-09-01
[NRLF-693] sonarcube fixes

nomad3k:


layer/psycopg2/psycopg2/pool.py
```diff
@@ -123,7 +123,7 @@ def _putconn(self, conn, key=None, close=False):
 
         # here we check for the presence of key because it can happen that a
         # thread tries to put back a connection after a call to close
-        if not self.closed or key in self._used:
+        if key in self._used:
```
jaklinger, line 126: why this this could have a weird side-effect as this is all just ported from [here](https://pypi.org/project/aws-psycopg2/) and so is a quasi-external dependency. possible to add a comment to exclude from sonarcube?

nomad3k: The lines above check for self.closed

jaklinger: yeah ok sounds legit, unless there's some weird side-effect between lines 100 and 122

jaklinger (approved):

PR #534, opened 2024-01-18
NRL-483 Fix all the SSO things

Tomdango:


scripts/_bootstrap.sh
```diff
@@ -72,10 +72,7 @@ function _bootstrap() {
     ;;
     #----------------
     "create-non-mgmt")
-      if [[ "$(aws sts get-caller-identity)" == *mgmt* ]]; then
-          echo "Please log in as a non-mgmt account" >&2
-          return 1
-      fi
+      _check_mgmt || return 1
```
Tomdango, line 75: Should be _check_not_mgmt

PR #551, opened 2024-03-26
NRL-578 Reuse DynamoDB resource and table

Tomdango:


mattdean3-nhs (approved):
Code looks good to me, works a treat and faster unit tests  :heavy_check_mark:  :heavy_check_mark: :heavy_check_mark:

Also, I'm getting a ~50ms response time on the producer read doc ref endpoint now (down from 150-200ms before) :partying_face:  

Nice work! :clap: :muscle: 

Could be worth updating the README with the new instructions or maybe worth a new ticket for this to review and update in bulk.

PR #556, opened 2024-04-04
[NRL-520] Move ASIDs to context.related for SSP content URLs

mattdean3-nhs:


layer/nrlf/core/validators.py
```diff
@@ -252,3 +262,59 @@ def _validate_relates_to(self, model: DocumentReference):
...
+        Validate that the document contains a valid ASID in the context.related field when the content contains an SSP URL
+        """
+
+        ssp_content = next(
+            filter(
+                _is_ssp_content,
+                model.content,
+            ),
+            None,
+        )
```
Tomdango:
```suggestion
        ssp_contents = any([
            content for content in model.content
            if _is_ssp_content(content)
        ])
```

layer/nrlf/core/validators.py
```diff
@@ -252,3 +262,59 @@ def _validate_relates_to(self, model: DocumentReference):
...
+
+        if not ssp_content:
+            logger.log(
+                LogReference.VALIDATOR001, step="ssp_asid", reason="no_ssp_content"
```
Tomdango: VALIDATOR001a

layer/nrlf/core/validators.py
```diff
@@ -252,3 +262,59 @@ def _validate_relates_to(self, model: DocumentReference):
...
+
+        if not getattr(model.context, "related", None):
+            self.result.add_error(
+                issue_code="required",
+                error_code="INVALID_RESOURCE",
+                diagnostics="Missing context.related. It must be provided and contain a valid ASID identifier when content contains an SSP URL",
+                field="context.related",
+            )
+            raise StopValidationError()
+
```
Tomdango, line 286:
```suggestion
    # Do comprehension check for >1 or 0
    
    for idx, asid_reference in enumerate(model.context.get("related", [])):
         # If not asid - continue
         # If asid - check is valid - return True/False depending on result
         
    # raise Exception here - we should never hit it
```

layer/nrlf/core/validators.py
```diff
@@ -252,3 +262,59 @@ def _validate_relates_to(self, model: DocumentReference):
...
+        logger.log(LogReference.VALIDATOR001, step="ssp_asid")
+
+        if not getattr(model.context, "related", None):
+            self.result.add_error(
+                issue_code="required",
+                error_code="INVALID_RESOURCE",
+                diagnostics="Missing context.related. It must be provided and contain a valid ASID identifier when content contains an SSP URL",
+                field="context.related",
+            )
+            raise StopValidationError()
```
Tomdango: Return instead

Tomdango (approved):
LGTM :+1:

PR #559, opened 2024-04-11
NRL-633 Persistent Environments for Pull Requests

Tomdango:
This adds new environments that get spun up when a pull request is opened.
The environment is updated every time the pull request is synchronised (updated).

When an environment is created or updated, a comment is added to the pull request with the latest deployed commit hash and the URL of the environment.

If an environment deploy fails, a comment should be added to the pull request with details of the run.

If the pull request is closed, the environment will be destroyed. **This will take precedence over any create/update jobs in progress and will cancel those jobs if running.**

If the pull request is reopened, the environment is recreated.


mattdean3-nhs (approved):
Worked a treat for me :clap: nice work!!!

Only thing I spotted was a warning in the action that a Terraform `output.json` file is not found but, given we're considering dropping the Terraform artifacts from the build, this should be fine left as is.

PR #564, opened 2024-04-12
NRL-641 Fix SonarCloud warnings

Tomdango:


api/producer/createDocumentReference/create_document_reference.py
```diff
@@ -105,71 +90,178 @@ def handler(
...
+            continue
+
+        _validate_pointer_details(existing_pointer, core_model, identifier, idx)
+
+        _append_id_if_replaces(relates_to, ids_to_delete, identifier)
+
+    return ids_to_delete
+
+
+def _validate_identifier(relates_to, idx):
```
eesa456, line 123: a lot of this logic is the same in the upsert endpoint too, i'd rather these new functions be moved to a common helpers file and used across both endpoints like that instead of duplicating it twice

api/producer/upsertDocumentReference/upsert_document_reference.py
```diff
@@ -101,61 +88,168 @@ def handler(
...
+            continue
+
+        _validate_pointer_details(existing_pointer, core_model, identifier, idx)
+
+        _append_id_if_replaces(relates_to, ids_to_delete, identifier)
+
+    return ids_to_delete
+
+
+def _validate_identifier(relates_to, idx):
```
eesa456, line 121: duplicated logic see previous comment

eesa456 (commented):
Looks really good, just some duplication for the upsert and create endpoints.

api/producer/createDocumentReference/create_document_reference.py
axelkrastek1-nhs: Good point. There were reasons why they where duplicated, but the upsert functionality seems to be here to stay for a while while 1D-sync is active.

axelkrastek1-nhs: Raised a ticket for this: https://nhsd-jira.digital.nhs.uk/browse/NRL-752

api/producer/upsertDocumentReference/upsert_document_reference.py
axelkrastek1-nhs: Raised a ticket for this: https://nhsd-jira.digital.nhs.uk/browse/NRL-752

api/producer/upsertDocumentReference/upsert_document_reference.py
```diff
@@ -104,70 +88,173 @@ def handler(
...
+                continue
+
+            _validate_pointer_details(existing_pointer, core_model, identifier, idx)
+
+            _append_id_if_replaces(relates_to, ids_to_delete, identifier)
+        except OperationOutcomeError:
+            if can_ignore_delete_fail:
+                logger.log(LogReference.PROUPSERT007f, relatesTo=resource.relatesTo)
+                continue
+            raise
```
eesa456: so before these errors returned spine error responses as 400, but now an operation outcome is raised in each method and then raised here again, would it not be better to return that spine error response that was used before. 
Also matt and I discussed this originally but if its a 1d sync request we dont want to do the existing pointer check and have to query the dynamo table twice as we are always gonna wanna delete with 1d sync anyways hence why it was in that if block so that we could skip that check.

api/producer/createDocumentReference/create_document_reference.py
```diff
@@ -105,71 +90,183 @@ def handler(
...
-                    identifier=identifier,
-                )
-                ids_to_delete.append(identifier)
-
-    if ids_to_delete:
+    for idx, relates_to in enumerate(resource.relatesTo):
+        try:
+            identifier = _validate_identifier(relates_to, idx)
+            _validate_producer_id(identifier, metadata, idx)
+            existing_pointer = _check_existing_pointer(identifier, repository, idx)
```
eesa456: same issue as other comment

api/producer/upsertDocumentReference/upsert_document_reference.py
eesa456: also in the repository we have this log REPOSITORY026a when we attempt the delete if it fails, we wont get this anymore with this logic. I'd prefer we skip the check existing pointer validation and go straight to the delete and catch it that way so we dont have to log and check here because querying the table using that gs1 index for every 1d sync request will slow down any get requests happening concurrently and we dont care about validating 1d sync requests as they already have been through NRL. just seems to be better optimised if we just do the delete and not query the dynamo and validate if the pointer exists and if it does exist thats more validation thats already been done in NRL which is taking more time. quicker to just go to the delete

axelkrastek1-nhs: Fair point. The exception does not get raised twice, it will only get raised once if it's not 1D-sync. If it is 1D-sync then it logs out that we are ignoring failures.

I thought it might be useful to understand better the cases when a failure is ignored in 1D-sync. But true that we might just go ahead and try to delete the pointer directly.

I'll do an early exit then for 1D-sync, instead of a big if

eesa456 (approved):

PR #585, opened 2024-04-18
NRL-516 add category validation

eesa456:


layer/nrlf/core/validators.py
```diff
@@ -316,3 +317,64 @@ def _validate_ssp_asid(self, model: DocumentReference):
...
+        """
+
+        if len(model.category) > 1:
+            logger.log(
+                LogReference.VALIDATOR001, step="category", reason="category_too_long"
+            )
+            self.result.add_error(
+                issue_code="invalid",
+                error_code="INVALID_RESOURCE",
+                diagnostics=f"Invalid category length: {len(model.category)}",
```
mattdean3-nhs: Could add info about what's expected too, maybe something like `Category must contain only a single value` or something similar.

layer/nrlf/core/validators.py
```diff
@@ -316,3 +317,64 @@ def _validate_ssp_asid(self, model: DocumentReference):
...
+            if coding.system != "http://snomed.info/sct":
+                self.result.add_error(
+                    issue_code="value",
+                    error_code="INVALID_RESOURCE",
+                    diagnostics=f"Invalid category system: {coding.system}",
+                    field=f"category[0].coding[{index}].system",
+                )
+                continue
+
+            if coding.code not in [
```
mattdean3-nhs: could create a global mapping for these codes, maybe something like this:

```
categories = [
  { 
    "code": "734163000",
    "display": "Care plan"
  }, 
  {
    "code": "1102421000000108",
    "display": "Observations"
  }
]
```

just a rough suggestion though, might bring the codes+displays together nicely. not saying you need to change anything here, I'll leave that up to you.

layer/nrlf/core/validators.py
```diff
@@ -316,3 +317,64 @@ def _validate_ssp_asid(self, model: DocumentReference):
...
+        logger.log(LogReference.VALIDATOR001, step="category")
+
+        logger.debug("Validating category")
+
+        for index, coding in enumerate(model.category[0].coding):
+            if coding.system != "http://snomed.info/sct":
+                self.result.add_error(
+                    issue_code="value",
+                    error_code="INVALID_RESOURCE",
+                    diagnostics=f"Invalid category system: {coding.system}",
```
mattdean3-nhs: Could add info about what's expected too, maybe add something like `Category system must be 'http://snomed.info/sct'`

layer/nrlf/core/validators.py
```diff
@@ -316,3 +317,64 @@ def _validate_ssp_asid(self, model: DocumentReference):
...
+                continue
+
+            if coding.code not in [
+                "734163000",
+                "1102421000000108",
+            ]:
+                self.result.add_error(
+                    issue_code="value",
+                    error_code="INVALID_RESOURCE",
+                    diagnostics=f"Invalid category code: {coding.code}",
```
mattdean3-nhs: Could add info about what's expected too, maybe something like `Category code must be '734163000' or '1102421000000108'.`

layer/nrlf/core/validators.py
eesa456: I originally wanted to do somat like this where I had the code as a key and the display as a value and then mapped the input code and display to the key value pair as validation but then thought that because theres only the 2 categories its not really necessary and unless we are expecting a lot more we dont really need to try and map as its just simpler to do it the current way but if we had a lot more categories then this would be a way better way of doing it than individually checking each one.  I might try and implement it to see if its quick tho

eesa456: alright improved the way its done i think lemme know what u think

mattdean3-nhs: Ah nice! I like that :muscle: :clap:

mattdean3-nhs (approved):
Nice, works a treat :heavy_check_mark: good work!

layer/nrlf/core/validators.py
```diff
@@ -316,3 +317,52 @@ def _validate_ssp_asid(self, model: DocumentReference):
...
+                    diagnostics=f"Invalid category system: {coding.system} Category system must be 'http://snomed.info/sct'",
+                    field=f"category[0].coding[{index}].system",
+                )
+                continue
+
+            if coding.code not in CATEGORIES.keys():
+                self.result.add_error(
+                    issue_code="value",
+                    error_code="INVALID_RESOURCE",
+                    diagnostics=f"Invalid category code: {coding.code} Category code must be '734163000' or '1102421000000108'.",
```
katebobyn-nhs: This is good but as we have at least four new categories planned in the near future, instead of needing to update this all the time we could just tell them what the Value Set is (because it will be dynamically updated). Something like, "Category must be a member of the England-NRLRecordCategory value set (https://fhir.nhs.uk/England/CodeSystem/England-NRLRecordCategory)"

mattdean3-nhs (approved):

PR #596, opened 2024-04-25
[NRL-669] Check if asid exists we still validate

eesa456:


layer/nrlf/core/validators.py
```diff
@@ -268,7 +268,21 @@ def _validate_ssp_asid(self, model: DocumentReference):
                 if content.attachment.url.startswith("ssp://")
             ]
         )
-        if not ssp_content:
+        does_related_exist = getattr(model.context, "related", None)
+        does_asid_exist = False
+        if does_related_exist:
+            asid_references = [
```
mattdean3-nhs, line 303: Functionally seems to work but I'm wondering if we could re-organise things a bit, to try and simplify the logic :thinking: 

Maybe something like:

1. Get SSP content
2. Get ASIDs
3. If ASIDs - validate that there's only one valid ASID
4. If SSP content but no ASID - add_error

If something like that would work, it would make the logic in this function simpler and would possibly minimise duplicated checks and such.

Just a suggestion mind, logic-wise the code is working for me.

layer/nrlf/core/tests/test_validators.py
```diff
@@ -722,6 +722,40 @@ def test_validate_ssp_content_without_any_context_related():
...
+                    "display": "Invalid identifier value",
+                }
+            ]
+        },
+        "diagnostics": "Invalid ASID value '1234'. context.related must contain a single valid ASID identifier when content contains an SSP URL",
+        "expression": ["context.related[0].identifier.value"],
+    }
+
+
+def test_validate_asid_with_no_ssp_content():
```
mattdean3-nhs: I don't think this test name represents the test, It's called `test_validate_asid_with_no_ssp_content` but the docref in the test does have SSP content but has no ASID.

layer/nrlf/core/tests/test_validators.py
```diff
@@ -722,6 +722,40 @@ def test_validate_ssp_content_without_any_context_related():
 
 
 def test_validate_ssp_content_without_asid_in_context_related():
```
mattdean3-nhs: I don't think this test name represents the test. It's called `test_validate_ssp_content_without_asid_in_context_related()` but the docref in the test doesn't have ssp content and does have an ASID but it's invalid.

mattdean3-nhs (commented):
I've left a few comments, happy to discuss things further if you'd like.

layer/nrlf/core/tests/test_validators.py
eesa456: i got the test names the wrong way round - oops ( the joys of copy paste)

layer/nrlf/core/validators.py
eesa456: made some changes to simplify the logic and reduce duplication

layer/nrlf/core/validators.py
```diff
@@ -318,6 +291,47 @@ def _validate_ssp_asid(self, model: DocumentReference):
...
+                (idx, related)
+                for idx, related in enumerate(getattr(model.context, "related", []))
+                if related.identifier.system == "https://fhir.nhs.uk/Id/nhsSpineASID"
+            ]
+            does_asid_exist = True
+            self._validate_asid(asid_references)
```
mattdean3-nhs: I think the new logic has introduced an unwanted validation error if you provide `context.related` but no ASID, you can verify it with this unit test:

```
def test_validate_with_context_related_but_no_asid():
    validator = DocumentReferenceValidator()
    document_ref_data = load_document_reference_json(
        "Y05868-736253002-Valid"
    )

    document_ref_data["context"]["related"] = [
        {
            "identifier": {
                "system": "https://fhir.nhs.uk/Id/ods-organization-code",
                "value": "Y05868",
            }
        }
    ]

    result = validator.validate(document_ref_data)

    assert result.is_valid is True
```

layer/nrlf/core/tests/test_validators.py
```diff
@@ -721,6 +739,40 @@ def test_validate_ssp_content_without_any_context_related():
...
+        "details": {
+            "coding": [
+                {
+                    "system": "https://fhir.nhs.uk/ValueSet/Spine-ErrorOrWarningCode-1",
+                    "code": "INVALID_IDENTIFIER_VALUE",
+                    "display": "Invalid identifier value",
+                }
+            ]
+        },
+        "diagnostics": "Invalid ASID value '1234'. context.related must contain a single valid ASID identifier when content contains an SSP URL",
```
mattdean3-nhs: Do we need to update this diagnostics message as this error no longer has anything to do with an SSP url?

layer/nrlf/core/validators.py
```diff
@@ -256,6 +256,32 @@ def _validate_relates_to(self, model: DocumentReference):
...
+            )
+            return
+
+        idx, asid_reference = asid_references[0]
+        asid_value = getattr(asid_reference.identifier, "value", "")
+        if not match(r"^\d{12}$", asid_value):
+            self.result.add_error(
+                issue_code="value",
+                error_code="INVALID_IDENTIFIER_VALUE",
+                diagnostics=f"Invalid ASID value '{asid_value}'. context.related must contain a single valid ASID identifier when content contains an SSP URL",
```
mattdean3-nhs: Maybe remove the SSP URL bit from this specific diagnostics message as it's no
 longer linked directly to SSP URLs.

layer/nrlf/core/validators.py
```diff
@@ -256,6 +256,32 @@ def _validate_relates_to(self, model: DocumentReference):
...
+        """
+        Validate that the ASID provided in the document is valid
+        """
+        logger.log(LogReference.VALIDATOR001, step="ssp_asid")
+
+        if len(asid_references) > 1:
+            self.result.add_error(
+                issue_code="invalid",
+                error_code="INVALID_RESOURCE",
+                diagnostics="Multiple ASID identifiers provided. context.related must contain a single valid ASID identifier when content contains an SSP URL",
```
mattdean3-nhs: Maybe remove the SSP URL bit from this specific diagnostics message as it's no longer linked directly to SSP URLs.

mattdean3-nhs (commented):
I think there's a few diagnostics messages to update to reflect having ASIDs without SSP URLs. See comments.

layer/nrlf/core/validators.py
eesa456: my only issue with this is that technically the context.related only does need a valid asid when there's SSP content outside of that that it doesnt but we wanna say that if there is an asid it needs to be valid not that it always needs to be provided idk this probably needs better rewording

eesa456: i changed it to what i think is more accurate for this circumstance

mattdean3-nhs: Yeah, I think that's bang on :clap:

mattdean3-nhs (approved):
:+1:

PR #602, opened 2024-05-01
[NRL-702] Add coding length validation 

eesa456:


api/producer/swagger.yaml
```diff
@@ -310,9 +310,9 @@ paths:
             ]
           }
           ```
-        * `category` SHOULD indicate the broader class of the Document Type as agreed during [onboarding](#api-description__onboarding). e.g.
+        * `category` MUST indicate the broader class of the Document Type as agreed during [onboarding](#api-description__onboarding). e.g.
           ```
-          "type": {
+          "category": {
```
katebobyn-nhs, line 315: Actually 'category', unlike type, is supposed to be in list format (even though we are only allowing one)... (also I realise this wasn't mentioned when discussing the ticket, I should have caught it earlier sorry).

```suggestion
          "category": { [
            "coding": [
              {
                "system": "http://snomed.info/sct",
                "code": "1102421000000108",
                "display": "Observations"
              }
            ]
          }
         ]
```

katebobyn-nhs (commented):
Might be worth checking with Hamza whether the same happens for 'type' as well, since anything with a coding can have multiple codings in theory (in FHIR). If that's the case it might be appropraite to pull out that logic so it can be used for both (and for any future similar coding arrays). 

The documentation changes look good except for my suggested amendment (we need the square braces); I think we might be doing these directly in the producer and consumer repositories now. Not sure whether the versions here need to be kept up to date or removed or ignored.

api/producer/swagger.yaml
eesa456: Type gets validated when trying to create the core model in the interaction using the DocumentPointer from the model class. So if it has multiple types or codings it throws an error.

katebobyn-nhs (approved):

PR #612, opened 2024-05-08
NRL-716 role arn var

eesa456:


.github/workflows/pr-env-destroy.yml
```diff
@@ -84,6 +84,7 @@ jobs:
           terraform -chdir=terraform/infrastructure destroy \
             --var-file=etc/dev.tfvars \
             --var assume_account=${{ steps.get_account_id.outputs.aws_account_id }} \
```
mattdean3-nhs: I think you only need the `role_arn` var now so `-var assume_account=` and `-var assume_role=` can be removed also. However, not needed for the fix the to work, it's just tidy up.

mattdean3-nhs (approved):
LGTM, nice work :clap: - just one comment about a possible tidy up to remove the old vars.

.github/workflows/pr-env-destroy.yml
eesa456: removed

.github/workflows/pr-env-destroy.yml
```diff
@@ -83,8 +83,7 @@ jobs:
         run: |
           terraform -chdir=terraform/infrastructure destroy \
             --var-file=etc/dev.tfvars \
-            --var assume_account=${{ steps.get_account_id.outputs.aws_account_id }} \
-            --var assume_role=terraform \
+            --var assume_role_arn=${{ secrets.DEPLOY_ROLE_ARN }} \
             -auto-approve
```
Tomdango, line 88: Suggestion - Could also clean up the Terraform workspace

```suggestion
      - name: Cleanup Terraform Workspace
        run: |
          terraform workspace select default
          terraform workspace delete ${{ needs.set-environment-id.outputs.environment_id }}

```

```

Tomdango (approved):
Approved - but with one possible improvement attached

.github/workflows/pr-env-destroy.yml
eesa456: added

Tomdango (approved):
:+1: LGTM

PR #614, opened 2024-05-09
[NRL-718] Always preserve original date field during pointer updates

mattdean3-nhs:


axelkrastek1-nhs (approved):
LGTM, I would have liked to see one feature test, but that could be done in a separate ticket.

PR #624, opened 2024-05-20
Feature/kabo5 nrl 525 self links serch

katebobyn-nhs:


api/consumer/searchPostDocumentReference/search_post_document_reference.py
```diff
@@ -61,9 +66,20 @@ def handler(
         if body.custodian_identifier
         else None
     )
+    if custodian_id:
+        self_link += f"&custodian:identifier=https://fhir.nhs.uk/Id/ods-organization-code|{custodian_id}"
```
mattdean3-nhs, line 70: Do the contents of the query string need URL encoding? I think there are a few characters in these strings that would usually require encoding, eg: colons, slashes, pipes.

tests/features/consumer/searchDocumentReference-success.feature
```diff
@@ -19,6 +19,7 @@ Feature: Consumer - searchDocumentReference - Success Scenarios
       | subject   | 9278693472 |
     Then the response status code is 200
     And the response is a searchset Bundle
+    And the Bundle has a self link matching consumer/FHIR/R4/DocumentReference?subject:identifier=https://fhir.nhs.uk/Id/nhs-number|9278693472"
```
mattdean3-nhs: is it worth commenting this out until after the integration test ticket is done? that way we can merge this ticket as-is and pull the integration tests in once they're ready?

api/consumer/searchPostDocumentReference/search_post_document_reference.py
```diff
@@ -45,6 +46,10 @@ def handler(
             expression="subject:identifier",
         )
 
+    config = Config()
+    base_url = f"https://{config.ENVIRONMENT}.api.service.nhs.uk/"
```
mattdean3-nhs, line 50: Do we need to support internal hostnames as well as the public ones? Environments that only have an internal hostname, like dev envs and integ tests etc will likely have invalid link urls as they have no public hostnames.

api/consumer/searchPostDocumentReference/search_post_document_reference.py
katebobyn-nhs: After discussion we've decided that the primary purpose these self-links serve atm is to show which search parameters were actually applied (the previous and next page links would be needed for pagination, which is not in scope for MVP; no consumer should need to follow this particular URL as it would take them to the exact result set they are currently reading). It might be more legible without URL encoding. We have agreed to consider URL encoding when pagination links are added to the array.

api/consumer/searchPostDocumentReference/search_post_document_reference.py
katebobyn-nhs: This is true. For now, at least public facing urls will always be valid. We could consider adding some logic to determine whether the internal format needs to be used instead, based on the environment in use, but that will not add any immediate value for users/suppliers at this point so it has been agreed to revisit this after MVP. We might also want to switch to relative URLs, which would be valid per FHIR spec since the link's `.url` is a "uri" type: https://hl7.org/fhir/R4/datatypes.html#uri

tests/features/consumer/searchDocumentReference-success.feature
mattdean3-nhs: nice fix, even better :clap:

mattdean3-nhs (approved):
LGTM :ballot_box_with_check:

PR #626, opened 2024-05-22
[NRL-604] Add K6 tests for producer stress scenarios

mattdean3-nhs:
Main differences are:
- Add K6 tests for producer stress test scenarios
- Change perf test prepare in `environment.py` to save entire pointer into `reference-data.json` file 
- Restructure K6 test tools to support creating/updating pointers

### DIFF
--- a/tests/performance/producer/stress.js
+++ b/tests/performance/producer/stress.js
@@ -1,0 +1,75 @@
+export * from "./client.js";
+
+export const options = {
+  tlsAuth: [
+    {
+      cert: open(`../../../truststore/client/${__ENV.ENV_TYPE}.crt`),
+      key: open(`../../../truststore/client/${__ENV.ENV_TYPE}.key`),
+    },
+  ],
+  scenarios: {
+    createDocumentReference: {
+      exec: "createDocumentReference",
+      executor: "ramping-vus",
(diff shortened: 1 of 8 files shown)


tests/performance/producer/stress.js
```diff
@@ -0,0 +1,80 @@
...
+    upsertDocumentReference: {
+      exec: "upsertDocumentReference",
+      executor: "ramping-vus",
+      startVUs: 1,
+      stages: [
+        { target: 10, duration: "30s" },
+        { target: 10, duration: "1m" },
+      ],
+    },
+    /*searchDocumentReference: {
```
eesa456: I'd rather these just be deleted if they're not going to be uncommented in the near future

eesa456 (approved):
Looks good, one small comment, feel free to ignore if you don't think its needed

tests/performance/producer/stress.js
mattdean3-nhs: yeah, good shout, I was on the fence about this but happy to remove as there's no guarantee these will be re-enabled in the short term (we need to fix the endpoints and the tests first)

eesa456 (approved):

PR #631, opened 2024-05-28
NRL-638 wip sample integration tests for producer transactions

katebobyn-nhs:


eesa456 (approved):

tests/features/steps/1_setup.py
```diff
@@ -62,4 +63,10 @@ def create_document_reference_step(context: Context):
...
 
     context.repository.create(doc_pointer)
-    context.add_cleanup(lambda: context.repository.delete(doc_pointer))
+    context.add_cleanup(clean_up_test_pointer, context, doc_pointer)
+
+
+def clean_up_test_pointer(context: Context, doc_pointer: DocumentPointer):
+    """Remove a pointer during cleanup without failing if it has already been deleted"""
+    with suppress(Exception):
+        context.repository.delete(doc_pointer)
```
axelkrastek1-nhs, line 72: It'd be interesting to know why this happens though?

tests/features/steps/2_request.py
```diff
@@ -82,9 +85,53 @@ def create_post_document_reference_step(context: Context, ods_code: str):
...
     if context.response.status_code == 201:
         doc_ref_id = context.response.headers["Location"].split("/")[-1]
+        doc_ref_id.replace(
+            "|", "."
+        )  # NRL-766 define and resolve custodian suffix behaviour
         context.add_cleanup(lambda: context.repository.delete_by_id(doc_ref_id))
 
 
+@when(
+    "producer 'TSTCUS' requests creation of a DocumentReference with default test values except '{section}' is"
```
axelkrastek1-nhs, line 95: Why is this not configurable like the one below. i.e: `{ods_code}`?

tests/features/producer/createDocumentReference-success.feature
```diff
@@ -1,23 +1,141 @@
-# Create care plan document reference
 Feature: Producer - createDocumentReference - Success Scenarios
 
-  Scenario: Successfully create a Document Pointer (care plan)
+  # Scenario: Successfully create a Document Pointer (care plan)
```
axelkrastek1-nhs: Why was this scenario commented out?

tests/features/producer/deleteDocumentReference-failure.feature
```diff
@@ -1,5 +1,44 @@
...
+      | id          | OC84-111-DeleteTest-NotYourPointer |
+      | subject     | 9278693472                         |
+      | status      | current                            |
+      | type        | 861421000000109                    |
+      | category    | 734163000                          |
+      | contentType | application/pdf                    |
+      | url         | https://example.org/my-doc.pdf     |
+      | custodian   | DK94                               |
+      | author      | N64                                |
+    When producer 'DK94' requests to delete DocumentReference with id 'OC84-111-DeleteTest-NotYourPointer'
```
axelkrastek1-nhs, line 22: I need a small discussion about this just to confirm custodians cannot delete their pointers even for cases when the author ODS code changes.

tests/features/producer/deleteDocumentReference-success.feature
```diff
@@ -1,3 +1,69 @@
 # Delete by pointer ID
-# Delete by pointer ID - URL encoded
-# Delete by pointer ID - not found
+Feature: Producer - deleteDocumentReference - Success Scenarios
+
+  Scenario: Successfully delete a Document Pointer by logical ID
+    Given the application 'DataShare' (ID 'z00z-y11y-x22x') is registered to access the API
```
axelkrastek1-nhs, line 5: Interesting that they can delete pointers types they are not authorised for. I guess it makes sense as they couldn't have created it in the first place. But maybe it is better to do a happy path case where they are authorised and this is for deleting any pointer with their custodian even if they are not authorised.

tests/features/producer/upsertDocumentReference-success.feature
```diff
@@ -1,3 +1,42 @@
+# Invalid target logical ID (wrong org prefix)
+# Invalid target logical ID (owned by this org, but pointer already exists)
+# Other edge cases will be same as create(POST)
+Feature: Producer - upsertDocumentReference - Success Scenarios
+
+  Scenario: Successfully create a Document Pointer with a chosen logical ID (care plan)
```
axelkrastek1-nhs, line 6: I think it'd be good if the logical ID could be set in the test, and add an assertion for this test that the ID is the one that was created.

tests/features/utils/data.py
```diff
@@ -73,4 +86,46 @@ def create_test_document_reference(items: dict) -> DocumentReference:
...
+      {custom_data if section=="subject" else DEFAULT_TEST_SUBJECT},
+      {custom_data if section=="custodian" else DEFAULT_TEST_CUSTODIAN},
+      {custom_data if section=="author" else DEFAULT_TEST_AUTHOR},
+      {custom_data if section=="type" else DEFAULT_TEST_TYPE},
+      {custom_data if section=="category" else DEFAULT_TEST_CATEGORY},
+      {custom_data if section=="description" else DEFAULT_TEST_DESCRIPTION},
+      {custom_data if section=="security" else DEFAULT_TEST_SECURITY_LABEL},
+      {custom_data if section=="content" else DEFAULT_TEST_CONTENT},
+      {custom_data if section=="context" else DEFAULT_TEST_CONTEXT}
+    }}"""
```
axelkrastek1-nhs, line 129: Nitpick, you can leave it as it is but you could have just modified here the relevant section.
```python
items = {
  ... populate with default values
}
doc_ref = create_test_document_reference(items)
doc_ref[section] = custom_data
```
`pointer_id` would become part of the items dictionary.

axelkrastek1-nhs (approved):

axelkrastek1-nhs:
There are some errors in the integration tests

tests/features/steps/1_setup.py
katebobyn-nhs: It happens when, for example, the intention is to delete the pointer. The "Delete" tests (happy path) delete the pointer that was created with this function. Without this change we get exceptions raised.

tests/features/steps/2_request.py
katebobyn-nhs: It can be, but I felt it would cause more confusion, beecause I wanted a helper method that would allow us to change one component to be non-compliant where _everything else_ is perfectly valid (so that we would know the component causing the failure would be the one we alter). 
The custodian changing would mean multiple components would become invalid for business logic reasons, since the producer ODS code sending the pointer must match the ODS code in multiple places in the body: custodian component, pointer ID and (if superceding) the context.relatesTo.

tests/features/producer/createDocumentReference-success.feature
katebobyn-nhs: accident, now fixed! thank you. Only the one below it (suffix) should have been.

tests/features/producer/deleteDocumentReference-failure.feature
katebobyn-nhs: > cases when the author ODS code changes.

You mean cases when the custodian ODS changes? The author changing shouldn't affect anything permissions wise

tests/features/utils/data.py
katebobyn-nhs: I thought about that first but then I wanted something where the whole document was just a string because failure scenario tests were requested where for example, there are duplicate keys in the request body, and my understanding was that a dictionary would just replace the key, or is there a way around that?

tests/features/steps/1_setup.py
katebobyn-nhs: raising ticket for better long-term solution (flags? separate setup step? cleanup steps at end of tests?)
+ to ensure different NHS numbers used for different feature suites so avoid interfering with subsequent search/count tests on failure to cleanup

tests/features/producer/deleteDocumentReference-failure.feature
katebobyn-nhs: Adding another case where target pointer has a different custodian
Results should be the same as pointer is never actually looked at if the ID mismatches custodian ODS

tests/features/producer/deleteDocumentReference-success.feature
katebobyn-nhs: Also will add a standard case where the permissions are current

axelkrastek1-nhs (approved):

PR #632, opened 2024-05-29
NRL-703 add perm check to upsert

eesa456:


api/producer/upsertDocumentReference/upsert_document_reference.py
```diff
@@ -127,27 +131,39 @@ def handler(
                 )
 
             if not (existing_pointer := repository.get_by_id(identifier)):
```
mattdean3-nhs: Could we optimize the upsert by only getting the existing pointer when the `PERMISSION_SUPERSEDED_IGNORE_DELETE_FAIL` permission isn't set? 

If that'd work, it would reduce the number of dynamodb queries for each superseded upsert call from 1dsync.

eesa456: but if we have the permission we still want to do that delete and get the existing pointer and check it right? if we have the permission and never check if the existing pointer is in NRLF then we'd wouldn't ever supersede anything, unless we never check the pointer exists and we just always do the delete for 1D sync and if it errors we catch it and ignore but that would still be us sending something to Dynamo? If a delete operation doesn't really take much time compared to a query then I wouldn't be opposed to doing it that way but i dont like the idea of willingly letting the delete fail and catching an exception and ignoring it just as a software developer principle 😂

mattdean3-nhs: yeah, was thinking the logic might be something like:

```
if PERMISSION_SUPERSEDED_IGNORE_DELETE_FAIL:
    - attempt to delete any superseded pointers but ignore (or just log) if delete fails
else:
    - read all superseded pointers from dynamodb and fail if any are missing
    - delete all existing pointers 
```

the main thing is that the read and delete are 2 separate operations per superseded pointer. we could potentially halve that overhead for 1dsync as we don't really care if the pointer exists or not.

Not saying we have to do this BTW, was just a thought as the logic ignores  `existing_pointer` if it's missing. If the validation of the superseded pointer is important for 1dsync, I'd be happy to leave this as it is now.

Also, the catching an exception from the delete failure will be needed anyway right? even if we've already read the existing pointer, we cannot guarantee that it'll still be there when the code gets to the delete.

eesa456: yeah i see what you mean, i know we carch the delete failure but we still log and return that as an error instead of ignoring, i dont really like the idea of letting something fail but in this case its alright. Another thing is if we say that the existing pointer should only be retrieved if we dont have the perm then we kinda need to switch up the checks for the existing pointer too to see if the perm exists as well, not a lot of work just a bit of switching stuff around, I'll see what i can come up with and if its worse I'll switch it back

mattdean3-nhs: yep, which also means that even if we keep the currently logic, we'll still need something to ignore when a delete fails for 1dsync.

mattdean3-nhs (approved):
LGTM - nice work :+1:

PR #638, opened 2024-06-04
NRL-738 allow sync full pointer access

eesa456:


layer/nrlf/core/decorators.py
```diff
@@ -66,7 +67,12 @@ def load_connection_metadata(headers: Dict[str, str], config: Config):
     logger.log(LogReference.HANDLER002, headers=headers)
     metadata = parse_headers(headers)
     logger.log(LogReference.HANDLER003, metadata=metadata.dict())
-
+    if metadata.nrl_permissions:
```
axelkrastek1-nhs: We should probably add an extra permission for this instead of only looking for the nrl_permissions.

axelkrastek1-nhs (approved):

PR #643, opened 2024-06-11
NRL-746 add extension to model

eesa456:


layer/nrlf/core/tests/test_validators.py
```diff
@@ -599,6 +599,195 @@ def test_validate_category_coding_invalid_system():
...
+                    "display": "Invalid validation of resource",
+                }
+            ]
+        },
+        "diagnostics": "Invalid content extension length: 2 Extension must only contain a single value",
+        "expression": ["content[0].extension"],
+    }
+
+
+def test_validate_content_extension_invalid_code():
```
axelkrastek1-nhs, line 643: What about when only one of them is invalid?

layer/nrlf/core/validators.py
```diff
@@ -116,6 +116,8 @@ def validate(self, data: Dict[str, Any] | DocumentReference):
             self._validate_relates_to(resource)
             self._validate_ssp_asid(resource)
             self._validate_category(resource)
+            if resource.content[0].extension:
```
axelkrastek1-nhs, line 119: Content is mandatory right? Do we check it somewhere? Otherwise if content is empty at this point, it would be a problem.

layer/nrlf/producer/fhir/r4/strict_model.py
```diff
@@ -358,11 +358,42 @@ class RequestHeaderCorrelationId(BaseModel):
     ]
 
 
+class CodeableConcept(BaseModel):
```
axelkrastek1-nhs, line 361: Why have this changes been added to the strict model in producer and not in consumer?

axelkrastek1-nhs (commented):
Final comment: This also means we should be updating the proxies/NRLF documentation with the new changes.

layer/nrlf/producer/fhir/r4/strict_model.py
eesa456: there is no strict model in the consumer folder

eesa456: but it is in the base model of consumer

layer/nrlf/core/validators.py
eesa456: we do validate the content exists against the model first i believe in our endpoints and also in the validate ssp asid method we check content stuff at that point too

eesa456: cos in the validate method we parse it into that document reference object from the pydantic model and if content isnt there it raises an exception as its mandatory in the model

layer/nrlf/core/tests/test_validators.py
eesa456: we dont check the display so it wont matter as long as the code is correct, but I can add in that check because code and display should be the same value anyways static or dynamic and add that test case

layer/nrlf/core/validators.py
axelkrastek1-nhs: if there is a unit test covering the case where the content is empty before this point, then this is ok.

layer/nrlf/producer/fhir/r4/strict_model.py
axelkrastek1-nhs: My bad, I thought they both had the strict_model. 😅

layer/nrlf/core/tests/test_validators.py
axelkrastek1-nhs: Hmm... maybe we should be checking for the display then? Probably in another ticket. But is there any reason why the display would be different than the expected value?

eesa456: added the new case and validation

eesa456: display would never be different in this case, validation is quite simple so ive just added it

axelkrastek1-nhs (approved):

layer/nrlf/core/validators.py
```diff
@@ -388,3 +390,107 @@ def _validate_category(self, model: DocumentReference):
...
+                logger.log(
+                    LogReference.VALIDATOR001,
+                    step=f"content[{i}].extension.valueCodeableConcept.coding[0].system",
+                    reason="extension_coding_system_invalid",
+                )
+                self.result.add_error(
+                    issue_code="value",
+                    error_code="INVALID_RESOURCE",
+                    diagnostics=f"Invalid content extension system: {content.extension[0].valueCodeableConcept.coding[0].system} Extension system must be 'https://fhir.nhs.uk/England/CodeSystem/England-NRLContentStability'",
+                    field=f"content[0].extension[0].valueCodeableConcept.coding[0].system",
```
axelkrastek1-nhs: Remove the f string here, as you are not replacing anything

eesa456: fixed it, it was a type should be i not 0

axelkrastek1-nhs (approved):

PR #650, opened 2024-06-19
WIP : NRL-737 add script and command

eesa456:


layer/nrlf/core/authoriser.py
```diff
@@ -53,3 +54,30 @@ def get_pointer_types(
...
+
+def parse_permissions_file(connection_metadata: ConnectionMetadata) -> list[str]:
+    ods_code = ".".join(connection_metadata.ods_code)
+
+    app_id = connection_metadata.client_rp_details.developer_app_id
+    ods_code = connection_metadata.ods_code
+    ods_code_extension = connection_metadata.ods_code_extension
+
+    if ods_code_extension:
+        key = f"{app_id}/{ods_code}.{ods_code_extension}.json"
```
axelkrastek1-nhs, line 67: This is ok, but it would be subject to change when we do NRL-766. So maybe a comment to remember would be good to have or a comment in the jira.

scripts/get_s3_permissions.py
```diff
@@ -0,0 +1,111 @@
...
+            else:
+                file_names.append(key)
+
+        next_token = response.get("NextContinuationToken")
+
+    return file_names, folders
+
+
+def download_files(s3_client, bucket_name, local_path, file_names, folders):
+    print(f"Downloading {len(file_names)} S3 files to temporary direrectory...")
```
axelkrastek1-nhs, line 76: There should be a check that you could parse each json file, and if not fail. We could raise a separate ticket for this.

eesa456: I feel like this check would make sense in the pipeline instead of here as this will happen on every release and make things longer but also all files in S3 should be valid JSON

eesa456: I also dont think we should fail if 1 file isnt parseable but all the others are we'd still wanna embed the others right for performance so maybe skip it?

axelkrastek1-nhs: Agreed. Then we should put it as acceptance criteria for the pipeline ticket 👍

axelkrastek1-nhs (approved):

PR #657, opened 2024-06-24
NRL-737 fix layer rewrite bug, change name and workflow

eesa456:


scripts/add-perms-to-lambda.sh
```diff
@@ -5,6 +5,15 @@ set -o errexit -o pipefail -o nounset
...
+mkdir -p "${dist_path}/nrlf_permissions-layer/python/nrlf_permissions"
+cp -r "${dist_path}/nrlf_permissions"/* "${dist_path}/nrlf_permissions-layer/python/nrlf_permissions"
+
+cd "${dist_path}/nrlf_permissions-layer/" && \
+    rm -f ${dist_path}/nrlf_permissions.zip && \
+    zip -q -r "${dist_path}/nrlf_permissions.zip" . \
+        -x **tests**\* **scripts**\* **__pycache__**\*
+
+rm -rf "${dist_path}/nrlf_permissions-layer"
+rm -rf "${dist_path}/nrlf_permissions"
```
mattdean3-nhs, line 17: As this script doesn't create this directory, it might be better to not delete it in this script

Not a biggie so happy to leave

terraform/infrastructure/layer.tf
```diff
@@ -10,7 +10,7 @@ module "third_party" {
   prefix = local.prefix
 }
 
-module "s3" {
+module "nrlf_permissions" {
```
mattdean3-nhs, line 13: much better name :clap: :muscle:

.github/workflows/pr-env-deploy.yml
```diff
@@ -78,6 +78,22 @@ jobs:
...
         run: make build
 
+      - name: Configure Management Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: eu-west-2
+          role-to-assume: ${{ secrets.MGMT_ROLE_ARN }}
+          role-session-name: github-actions-ci-${{ needs.set-environment-id.outputs.environment_id }}
+
+      - name: Setup Python environment
```
mattdean3-nhs: Do we need to run the "Setup Python environment" step again? It's already run earlier in this action :point_up:

mattdean3-nhs (approved):
Nice work, works a treat :muscle: 

A few minor comments but nothing that's urgent so am approving.

scripts/add-perms-to-lambda.sh
eesa456: true but it does use it to copy the downloaded files across so i cant delete it in the python script without causing an issue, it is all in the dist anyways so even if i didnt remove it wouldnt stay so its fine

.github/workflows/pr-env-deploy.yml
eesa456: yeah copy paste has exposed me 😂

mattdean3-nhs (approved):

PR #671, opened 2024-07-02
NRL-791 add regex restrictions to model, disable custodian suffixes

katebobyn-nhs:


katebobyn-nhs:
not needed with update to pydantic2

PR #673, opened 2024-07-02
NRL-817 permissions cache

axelkrastek1-nhs:


mattdean3-nhs (approved):
:heavy_check_mark: Tested both persistent-env and pr deploys, both worked and had the correct permissions when deployed. The build-artifacts zip file also no longer contains the permissions.

Nice work :clap:

PR #684, opened 2024-07-08
NRL-846 Use the auth store policy for lambdas

axelkrastek1-nhs:


mattdean3-nhs (approved):
LGTM - works well too :muscle:

PR #696, opened 2024-07-17
NRL-820 add new pointer types to constants

eesa456:


terraform/account-wide-infrastructure/modules/lambda-errors-metric-alarm/iam.tf
```diff
@@ -13,9 +13,43 @@ resource "aws_iam_policy" "lambda-errors-topic-kms-read-write" {
...
+
+data "aws_iam_policy_document" "sns_kms_key_policy" {
+  policy_id = "CloudWatchEncryptUsingKey"
+
+  statement {
+    effect = "Allow"
+    actions = [
+      "kms:*"
+    ]
+    resources = ["*"]
```
axelkrastek1-nhs, line 34: Is there a way to make this more specific?

layer/nrlf/core/constants.py
```diff
@@ -66,6 +68,8 @@ class Categories(Enum):
     PointerTypes.EOL_CARE_PLAN.value: Categories.CARE_PLAN.value,
     PointerTypes.LLOYD_GEORGE_FOLDER.value: Categories.CARE_PLAN.value,
     PointerTypes.NEWS2_CHART.value: Categories.OBSERVATIONS.value,
+    PointerTypes.ADVANCED_CARE_PLAN.value: Categories.CARE_PLAN.value,
+    PointerTypes.TREATMENT_ESCALATION_PLAN.value: Categories.CARE_PLAN.value,
```
axelkrastek1-nhs, line 72: It would be good to have a test that fails when a new pointer type gets added without the mapping to the category.

axelkrastek1-nhs: For reference this is where it would fail: https://github.com/NHSDigital/NRLF/blob/6adbb5a9f10b7564720b3bf17497f1c3081dc185/layer/nrlf/core/dynamodb/repository.py#L20

terraform/account-wide-infrastructure/modules/lambda-errors-metric-alarm/iam.tf
eesa456: This enables IAM user permissions for the key so we kinda need it to be able to do stuff with this key, idk if making it more specific to sns and cloudwatch only work work because we need it to work with cloudwatch as a service as oppose to just that alarm

layer/nrlf/core/constants.py
eesa456: we dont currently have any tests for the repository file (not sure why tbh) I could add this test as a start ? or make it as part of another ticket ?

eesa456: added some tests

axelkrastek1-nhs (approved):
Well done Eesa! ✅

PR #697, opened 2024-07-24
NRL-756 log retention periods

eesa456:


layer/nrlf/core/validators.py
```diff
@@ -222,7 +222,7 @@ def _validate_relates_to(self, model: DocumentReference):
         """"""
         if not model.relatesTo:
             logger.log(
-                LogReference.VALIDATOR001, step="relates_to", reason="no_relates_to"
```
mattdean3-nhs, line 225: How come we're changing all these validator logs as part of this PR? (it's not mentioned in the ticket)

terraform/infrastructure/etc/prod.tfvars
```diff
@@ -1,4 +1,5 @@
-account_name        = "prod"
-domain              = "api.record-locator.national.nhs.uk"
-public_domain       = "api.service.nhs.uk"
-deletion_protection = true
+account_name         = "prod"
+domain               = "api.record-locator.national.nhs.uk"
+public_domain        = "api.service.nhs.uk"
+deletion_protection  = true
+log_retention_period = 2192
```
mattdean3-nhs, line 5: I think this is 6 years + 2 days, is that to account for leap years? :brain:

terraform/infrastructure/etc/dev.tfvars
```diff
@@ -3,3 +3,4 @@ account_name = "dev"
 domain                = "api.record-locator.dev.national.nhs.uk"
 public_domain         = "internal-dev.api.service.nhs.uk"
 public_sandbox_domain = "internal-dev-sandbox.api.service.nhs.uk"
+log_retention_period  = 90
```
mattdean3-nhs, line 6: 90 days seems like a lot of logs for a dev env but happy to leave as is and review volume/cost over time.

terraform/infrastructure/etc/qa.tfvars
```diff
@@ -3,3 +3,4 @@ account_name = "qa"
 domain                = "qa.record-locator.national.nhs.uk"
 public_domain         = "internal-qa.api.service.nhs.uk"
 public_sandbox_domain = "internal-qa-sandbox.api.service.nhs.uk"
+log_retention_period  = 90
```
mattdean3-nhs, line 6: as with dev, seems like a long period for a qa env. Happy to review over time though.

terraform/infrastructure/vars.tf
```diff
@@ -43,3 +43,8 @@ variable "use_shared_resources" {
   type    = bool
   default = false
 }
+
+variable "log_retention_period" {
```
mattdean3-nhs, line 47: Is it worth highlighting that the value of this is in days? it might help future us so we don't have to dig deep into the TF code to confirm it.

A few options that sprint to mind are maybe change the var name to `log_retention_days` or maybe add a `description` to the var.

Not a major thing though so happy if this stays as is.

layer/nrlf/core/validators.py
eesa456: was going through the logs during the documentation stuff and realised that it was the wrong one so just made a few fixes

terraform/infrastructure/etc/prod.tfvars
eesa456: yeah, cloudwatch has certain numbers as well you can choose from too

terraform/infrastructure/etc/dev.tfvars
eesa456: yeah thats what it was agreed on

terraform/infrastructure/vars.tf
eesa456: feel like its a bit obvious from the fact that the values are super high like 2192 months is crazy 😂

mattdean3-nhs: :+1: yeah, not a major thing.

mattdean3-nhs (approved):

PR #701, opened 2024-08-14
[NRL-972] Update Pydantic to v1.10.16. Switch lambdas to FunctionUpdate mode for runtime updates

mattdean3-nhs:


hamzakhan24-nhs (approved):
Matt went through the changes with myself and Aud. Approving on the basis that this is an emergency hotfix with minimal changes that are just a version bump to pydantic and lambda config change for AWS to prevent auto update to the runtime versions. Will be tested in non prod environments after merge.

PR #704, opened 2024-08-27
NRL-973 Add headers to Performance Tests and Fail Workflow if Tests Fail

eesa456:


tests/performance/consumer/client.js
```diff
@@ -24,6 +26,13 @@ function getHeaders(odsCode = ODS_CODE) {
   };
 }
 
+function checkResponse(res) {
+  const is_success = check(res, { "status is 200": (r) => r.status === 200 });
+  if (!is_success) {
+    console.log(res.json());
```
mattdean3-nhs: ran the tests, and seeing 10% issues (probs my env rather than this PR). The issues look like this in the console:
```
INFO[0002] {"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"forbidden","details":{"coding":[{"system":"https://fhir.nhs.uk/ValueSet/Spine-ErrorOrWarningCode-1","code":"ACCESS DENIED","display":"Access has been denied to process this request"}]},"diagnostics":"The requested DocumentReference is not of a type that this organisation is allowed to access"}]}  source=console
```

Is it worth setting this to be `console.warn()` instead of `console.log()` to highlight that something unexpected has happened?

mattdean3-nhs (dismissed):
LGTM :heavy_check_mark: Nice work :clap: 

One comment on the log level when we get an unexpected response but it's no biggie and not important for the actual fix.

tests/performance/consumer/client.js
eesa456: not seen any issues myself on the tests for this PR at least but changed to warn

mattdean3-nhs (approved):
:heavy_check_mark:

PR #707, opened 2024-09-11
[NRL-863] Add version to env config

mattdean3-nhs:


scripts/set_env_config.py
```diff
@@ -0,0 +1,45 @@
...
+    if parameter_name in parameters:
+        current_value = parameters[parameter_name]
+        if current_value == parameter_value:
+            print(
+                f"'{parameter_name}' already set to '{parameter_value}'. No changes made."
+            )
+            sys.exit(0)
+
+        print(
+            f"Updating '{parameter_name}' from '{current_value}' to '{parameter_value}'...."
```
axelkrastek1-nhs, line 26: It would be good to get the information about the active stack. I think this print only shows the version of the inactive stack?

.github/workflows/persistent-environment.yml
```diff
@@ -232,6 +232,11 @@ jobs:
       - name: Terraform Apply
         run: terraform -chdir=terraform/infrastructure apply tfplan
 
+      - name: Update environment config version
+        run: |
+          deployed_version="${{ inputs.branch_name }}"
```
axelkrastek1-nhs: Not an issue, but an observation that we should always be using tags rather than making builds from develop. Alternatively, we could always append a short commit hash for those cases.

scripts/set_env_config.py
mattdean3-nhs: Yeah, it only shows the details of the parameter that is being updated, not the entire config.

Are you thinking it'd be good to show the entire config somewhere in the build logs? If so, I'll have a look and see if I can get that added (if it's not already there).

.github/workflows/persistent-environment.yml
mattdean3-nhs: I think it's fair to say the current deployed version is `develop` but could maybe do with the commit hash added in there too..... lemme have a look at that, maybe something like `{branch}@{short-commit-hash}` would be good for that case. What do you reckon?

scripts/set_env_config.py
mattdean3-nhs: Commented offline - going to confirm that we already have this and if not, add it :eyes:

.github/workflows/persistent-environment.yml
mattdean3-nhs: Confirmed offline - going to add this :eyes:

axelkrastek1-nhs (approved):

scripts/set_env_config.py
mattdean3-nhs: Confirmed that the before-and-after of the env config is already shown in the github action logs: 
Screenshot from 2024-09-17 10-11-34

PR #710, opened 2024-09-18
Feature/hakh11 nrl 1002 consumer integration tests

hamzakhan24-nhs:
**-Refactored and added scenarios to two feature files:**
 SearchPostDocumentReference-success
 SearchDocumentReference-success
 **-Added "custom tag" MakeFile command to enable selective scenario runs**


tests/features/consumer/searchDocumentReference-success.feature
```diff
@@ -36,75 +36,42 @@ Feature: Consumer - searchDocumentReference - Success Scenarios
...
+      | type        | 736253002                       |
+      | category    | 734163000                       |
+      | contentType | application/pdf                 |
+      | url         | https://example.org/my-doc.pdf  |
+      | custodian   | 02V                             |
+      | author      | 02V                             |
     When consumer 'RX898' searches for DocumentReferences with parameters:
       | parameter | value      |
       | subject   | 9278693472 |
+      | custodian | 02V        |
```
katebobyn-nhs, line 58: As far as I can tell, tests a search using both NHS# + custodian filters. If there is only one pointer and it has a matching NHS# + custodian, in theory the scenario would pass even if either the subject filter or the custodian filter were not working. It might be good to additionally set up a few that we would expect to be filtered out: one with matching custodian but wrong NHS number, and vice versa.

tests/features/consumer/searchPostDocumentReference-success.feature
```diff
@@ -174,6 +211,34 @@ Feature: Consumer - searchDocumentReference - Success Scenarios
...
+      | system                 | value           |
+      | http://snomed.info/sct | 736253002       |
+      | http://snomed.info/sct | 861421000000109 |
+    And the organisation 'RX898' is authorised to access pointer types:
+      | system                 | value           |
+      | http://snomed.info/sct | 736253002       |
+      | http://snomed.info/sct | 861421000000109 |
+    And a DocumentReference resource exists with values:
+      | property    | value                          |
+      | id          | DK94-111-DeleteDocRefTest1     |
```
katebobyn-nhs, line 226: is this a deletion test, if so does it belong in the delete feature? I see that it searches after the deletion so it's more of a combined scenario, this could be reflected in the scenario title if it lives here (it's slightly more complex than simply searching by NHS number where there never were any pointers for that patient)

Makefile
```diff
@@ -86,6 +86,14 @@ test-features-integration: check-warn ## Run the BDD feature tests in the integr
 		--define="use_shared_resources=${USE_SHARED_RESOURCES}" \
 		$(FEATURE_TEST_ARGS)
 
+integration-test-with-custom_tag:
+	@echo "Running feature tests in the integration environment ${TF_WORKSPACE_NAME}"
+	behave --define="integration_test=true" --tags=@custom_tag \
+		--define="env=$(TF_WORKSPACE_NAME)" \
+		--define="account_name=$(ENV)" \
+		--define="use_shared_resources=${USE_SHARED_RESOURCES}" \
+		$(FEATURE_TEST_ARGS)
```
katebobyn-nhs, line 95: ! this seems very handy, could you add a line to the readme specifying this option alongside where it mentions the full integration test run?

tests/features/consumer/searchDocumentReference-success.feature
hamzakhan24-nhs: have added two new scenarios which isolate each filter:

Search for a DocumentReference by NHS Number and Custodian where NHS number filter matches one pointer from result set

Search for a DocumentReference by NHS Number and Custodian where Custodian filter matches one pointer from result set

tests/features/consumer/searchPostDocumentReference-success.feature
hamzakhan24-nhs: Perhaps the deletion is not required, but the main goal of this test is to setup an NHS number which has 0 pointers, so I figured one way to do that is to create a pointer for an NHS number and then delete that pointer. May not be necessary.

Makefile
hamzakhan24-nhs: Readme now updated 👍

katebobyn-nhs (approved):

PR #718, opened 2024-10-09
NRL-1049 update allowed categories and types for imaging

eesa456:
…it can search by new NICIP type

### DIFF
--- a/resources/fhir/NRLF-RecordType-ValueSet.json
+++ b/resources/fhir/NRLF-RecordType-ValueSet.json
@@ -68,6 +68,14 @@
         {
           "code": "2181441000000107",
           "display": "Personalised Care and Support Plan"
+        },
+        {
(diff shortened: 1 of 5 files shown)


resources/fhir/NRLF-RecordType-ValueSet.json
```diff
@@ -68,6 +68,14 @@
...
           "code": "2181441000000107",
           "display": "Personalised Care and Support Plan"
+        },
+        {
+          "code": "MAULR",
+          "display": "MRA Upper Limb Rt"
+        },
+        {
+          "code": "MAXIB",
+          "display": "MRI Axilla Both"
```
katebobyn-nhs: These are under the SNOMED system like this, they need to be under NICIP

mattdean3-nhs (approved):
nice work! :clap:

PR #722, opened 2024-10-24
NRL-1132 flaky smoke test

eesa456:


mattdean3-nhs (approved):
Nice fix! Super tiny change and works really well :muscle:

PR #724, opened 2024-11-04
NRL-474 validate author system, length and value format

eesa456:


layer/nrlf/core/tests/test_validators.py
```diff
@@ -599,22 +599,16 @@ def test_validate_category_coding_invalid_system():
     }
 
 
-def test_validate_content_extension_too_many_extensions():
+def test_validate_author_too_many_authors():
```
mattdean3-nhs, line 643: Did you mean to remove the `test_validate_content_extension_too_many_extensions()` unit test?

mattdean3-nhs (commented):
All looks good and appears to work well :muscle: :clap: 

Just one comment from me on the removed unit test

layer/nrlf/core/tests/test_validators.py
eesa456: nope

mattdean3-nhs (approved):

PR #726, opened 2024-11-08
NRL-1050 Validate types and expand category validation

axelkrastek1-nhs:


layer/nrlf/core/tests/test_validators.py
```diff
@@ -485,6 +502,7 @@ def test_validate_category_coding_display_mismatch_observations():
     }
 
 
+# TODO: FIX name of this test to multiple codings
```
eesa456: any reason this hasn't been done already?

layer/nrlf/core/tests/test_validators.py
```diff
@@ -531,6 +549,7 @@ def test_validate_category_coding_invalid_code():
     }
 
 
+# TODO: FIX name of this test to invalid code
```
eesa456: same

layer/nrlf/core/tests/test_validators.py
```diff
@@ -599,6 +618,185 @@ def test_validate_category_coding_invalid_system():
...
+            {
+                "system": type_system,
+                "code": type_code,
+                "display": "some random display name",
+            }
+        ]
+    }
+
+    # match type to category to avoid getting that error
+    # TODO: Discuss in review if I should just ignore the error that happens to simplify test
```
eesa456: I don't think you should ignore the error personally, dont want a test having an error thats not expected

layer/nrlf/core/validators.py
```diff
@@ -389,7 +444,30 @@ def _validate_category(self, model: DocumentReference):
...
+        Validate the type field matches the expected category
+        """
+        logger.log(LogReference.VALIDATOR001, step="type_category_mapping")
+
+        type_coding = model.type.coding[0]
+        type_id = f"{type_coding.system}|{type_coding.code}"
+        category_coding = model.category[0].coding[0]
+        category_id = f"{category_coding.system}|{category_coding.code}"
+
+        if type_id not in PointerTypes.list() or category_id not in Categories.list():
```
eesa456, line 472: would this ever even happen considering previous validation?

layer/nrlf/core/tests/test_validators.py
```diff
@@ -413,6 +419,7 @@ def test_validate_category_too_many_category():
     }
 
 
+# TODO: Parameterize this test and the one below
```
eesa456: is this going to be a tech debt ?

layer/nrlf/core/tests/test_validators.py
axelkrastek1-nhs: Making the code review easier 🤣  I can change it once you had a complete view of the changes. Otherwise the diff is going to look ugly

layer/nrlf/core/tests/test_validators.py
axelkrastek1-nhs: same as: https://github.com/NHSDigital/NRLF/pull/726#discussion_r1841870340

layer/nrlf/core/tests/test_validators.py
axelkrastek1-nhs: 👍  I'll remove this comment

layer/nrlf/core/validators.py
axelkrastek1-nhs: Yes, it would happen, the previous validation is on another function that would exit early if there is an error and then come to this one.

layer/nrlf/core/tests/test_validators.py
axelkrastek1-nhs: I think I could do it right away, I didn't want to make the diff more complicated with non-related code, same as some comment above.

layer/nrlf/core/validators.py
eesa456: i see i guess then thats fine but if its not in our defined list of types or categories surely we'd want some log or something letting us know if not a error response

layer/nrlf/core/tests/test_validators.py
eesa456: fair enough i dont really mind if its not but if its not there should be some tech debt ticket for it

layer/nrlf/core/constants.py
```diff
@@ -113,6 +113,51 @@ def coding_value(self):
...
+TYPE_ATTRIBUTES = {
+    PointerTypes.MENTAL_HEALTH_PLAN.value: {
+        "display": "Mental health crisis plan",
+    },
+    PointerTypes.EMERGENCY_HEALTHCARE_PLAN.value: {
+        "display": "Emergency health care plan",
+    },
+    PointerTypes.EOL_COORDINATION_SUMMARY.value: {
+        "display": "End of Life Care Coordination Summary",
+    },
```
katebobyn-nhs, line 125: This is slightly different from https://github.com/NHSDigital/NRLF/blob/develop/resources/fhir/NRLF-RecordType-ValueSet.json, since it's case sensitive could this introduce problems? Is there a way to use the actual value set values here without duplicating them all manually?

layer/nrlf/core/validators.py
axelkrastek1-nhs: That already happens in the type/category validation, if they are not in that list, it would throw an error

layer/nrlf/core/constants.py
axelkrastek1-nhs: That's a great idea Kate, I fixed the case of the display and as I mention we can do the DRY change in another ticket.

eesa456 (approved):

PR #745, opened 2024-11-18
NRL-1177 Fix (again) pattern matching for sonar cloud exclusion

axelkrastek1-nhs:


mattdean3-nhs:
This one doesn't work

PR #754, opened 2024-11-29
add context to smoke test setup

eesa456:


.github/workflows/pr-env-deploy.yml
```diff
@@ -332,3 +332,53 @@ jobs:
 
       - name: Cleanup Environment Test Data
         run: make test-performance-cleanup TF_WORKSPACE_NAME=${{ needs.set-environment-id.outputs.environment_id }}
+
+  smoke-test:
```
mattdean3-nhs: :heart_eyes:

.github/workflows/pr-env-deploy.yml
```diff
@@ -332,3 +332,53 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: eu-west-2
+          role-chaining: true
+          role-to-assume: ${{ secrets.DEPLOY_ROLE_ARN }}
+          role-session-name: github-actions-ci-${{ needs.set-environment-id.outputs.environment_id }}
+
+      - name: Smoke Test
+        run: |
+          make TF_WORKSPACE_NAME=${{ needs.set-environment-id.outputs.environment_id }} test-smoke-public
```
mattdean3-nhs: this won't run against the PR env - `test-smoke-public` uses the public endpoint which will be `internal-dev`. I think you'll want `test-smoke-internal` instead.

mattdean3-nhs (approved):

PR #766, opened 2024-12-11
NRL-1202 validate type not just system

eesa456:


mattdean3-nhs (approved):
Nice :+1: :sunglasses:

PR #770, opened 2024-12-12
[NRL-1228] Fix advance care plan naming. Add report script to find invalid pointers

mattdean3-nhs:


katebobyn-nhs (approved):

reports/find_invalid_pointers.py
```diff
@@ -0,0 +1,81 @@
...
+    docref = DocumentReference.model_validate_json(document)
+
+    validator = DocumentReferenceValidator()
+    result = validator.validate(data=docref)
+
+    if not result.is_valid:
+        raise Exception("Failed to validate document: " + str(result.issues))
+
+
+def _find_invalid_pointers(table_name: str) -> dict[str, float]:
```
axelkrastek1-nhs, line 27: nitpick
```suggestion
def _find_invalid_pointers(table_name: str) -> dict[str, float | int]:
```

axelkrastek1-nhs (approved):
It looks good but we need to apply those changes in the proxy schema as well, and also sonarcloud warnings need fixing.

PR #781, opened 2024-12-19
NRL-786 new parent class that does not allow extra fields for pydantic models

eesa456:


layer/nrlf/core/parent_model.py
```diff
@@ -0,0 +1,88 @@
...
+        Optional[str],
+        Field(
+            description="A human language representation of the concept as seen/selected/uttered by the user who entered the data and/or which represents the intended meaning of the user.",
+            pattern="[ \\r\\n\\t\\S]+",
+        ),
+    ] = None
+
+
+class ParentExtension(BaseModel):
+    valueCodeableConcept: Annotated[
```
mattdean3-nhs, line 71: If we want to allow producers to define their own extensions, do we need `value` to be more generic to support extensions with values of types other than CodableConcepts?

eesa456: I'm not sure I've just kept the extension as its always been defined but could make it more generic ?

mattdean3-nhs: As discussed offline, we're going to support extensions that have a CodableConcept value only on the initial implementation. We will add a question to producer supplier onboarding to confirm if the pointers being created will have other custom extentions and then we'll extend our validation as required.

mattdean3-nhs (approved):

PR #801, opened 2025-01-14
NRL-1215 Detect duplicate fields

axelkrastek1-nhs:


.pre-commit-config.yaml
```diff
@@ -34,7 +34,7 @@ repos:
         additional_dependencies:
           - flake8-print
         args:
-          - "--select=T201,F401,F402,F403"
+          - "--select=F401,F402,F403"
```
axelkrastek1-nhs: This is only temporary for the prints, once I remove the prints, I'll remove this as well.

layer/nrlf/core/json_duplicate_checker.py
```diff
@@ -0,0 +1,125 @@
...
+    Tracks duplicate keys by maintaining path context during traversal.
+    Paths are recorded in dot notation with array indices:
+    - Objects: parent.child
+    - Arrays: parent.array[0]
+    - Nested: parent.array[0].child[1].key
+    """
+
+    def __init__(self):
+        # Here a list of paths because the same key name could be at different levels
+        self.duplicate_keys_and_paths: OrderedDict[str, list[str]] = OrderedDict()
```
axelkrastek1-nhs: It doesn't need to be an ordered dict. The problem is that when testing, the order is not ensured... So unit testing, but particularly integration tests would intermittently fail. Unit tests could be fixed by sorting the output which I do already, but that cannot be done for integration tests.

layer/nrlf/core/json_duplicate_checker.py
```diff
@@ -0,0 +1,125 @@
...
+            serialized_item = json.dumps(item, sort_keys=True)
+            if serialized_item in seen_elements:
+                element = f"{array_path}[{idx}]"
+                duplicate_path = ".".join(base_path + [element])
+                self.duplicate_keys_and_paths.setdefault(element, []).append(
+                    duplicate_path
+                )
+                print(f"Found duplicate array element at path: {duplicate_path}")
+            else:
+                seen_elements.add(serialized_item)
```
axelkrastek1-nhs: The code above was the most recent fix, it was not part of the ticket but I noticed that a duplicate element in an array wouldn't be flagged. Example:
```json
{ "array": ["duplicate", "duplicate"] }
```

mattdean3-nhs: a duplicate valued element in an array seems valid to me so I don't think we'd want it to failed like this - happy to discuss this one in more details if needed

mattdean3-nhs:
As discussed offline - I had a little play with using the `object_pairs_hook` parameter for `JSON.loads` for this. The idea being that we find the dupes as the JSON is loaded instead of loading it and then parsing it afterwards.

The code changes I made are on this branch: https://github.com/NHSDigital/NRLF/tree/feature/made14-NIR-1215-use-hooks-instead

I didn't have time to get all the unit tests passing before going on leave but I think the progress I made suggests (to me at least) that it's feasible to use `object_pairs_hook` to do what we want.

One to discuss - whether we put more time into trying to get a simpler solution _or_ we've spent enough time and this solution is the way we want to go.

layer/nrlf/core/json_duplicate_checker.py
```diff
@@ -0,0 +1,71 @@
+import json
+from typing import Any
+
+
+def check_for_duplicate_keys(pairs: list[tuple[str, Any]]) -> dict:
```
eesa456: this method and the one below both need a description i.e., """ this method does somat lol """

tests/features/producer/createDocumentReference-duplicateField.feature
```diff
@@ -0,0 +1,128 @@
...
+            {
+                "system": "https://fhir.nhs.uk/ValueSet/Spine-ErrorOrWarningCode-1",
+                "code": "MESSAGE_NOT_WELL_FORMED",
+                "display": "Message not well formed"
+            }
+            ]
+        },
+        "diagnostics": "Duplicate keys found in FHIR document: ['url']",
+        "expression": [
+            "root.content[0].attachment.url"
```
eesa456: i dont like the fact that it says root and not "entry." idk if thats somat that can be easily changed but all other error messages have entry. at the start

layer/nrlf/core/request.py
```diff
@@ -99,6 +101,22 @@ def parse_body(
         ) from None
 
 
+def raise_when_duplicate_keys(json_content: str) -> None:
+    """
+    Raises an error if duplicate keys are found in the JSON content.
+    """
+    logger.log(LogReference.HANDLER018)
+    duplicates, paths = check_duplicate_keys(json_content)
+    if duplicates:
```
eesa456, line 110: could add a log here saying that duplicates were found and what they might be - not too fussed but would match how other errors are raised in lambdas

layer/nrlf/core/tests/test_json_duplicate_checker.py
```diff
@@ -0,0 +1,332 @@
...
+            paths.append(f"{current_path}.array_obj.key")
+
+        return sorted(paths)
+
+    def test_parametrized_nested_arrays(self):
+        """Test different depths of nested arrays with duplicates at each level."""
+        for depth in range(1, 11):  # Test depths 1 through 10
+            with self.subTest(depth=depth):
+                json_content = self.generate_nested_json(depth)
+                print(f"\n=== Testing depth {depth} ===")
```
eesa456: this will get flagged

layer/nrlf/core/request.py
axelkrastek1-nhs: That's a good point, I'll put it in, thanks

layer/nrlf/core/json_duplicate_checker.py
axelkrastek1-nhs: done, thanks

tests/features/producer/createDocumentReference-duplicateField.feature
mattdean3-nhs: fixed - it now says `DocumentReference.` like the FHIR validator reports.

eesa456 (approved):

PR #841, opened 2025-02-19
Postman tests now contain new validation

audreygunson1:
The API catalogue postman tests now contain the new validation that was put in place eg format url and content stability segment


mattdean3-nhs (approved):
The collection works fine for me :heavy_check_mark: nice work!

The three pointer ids that you changed, those pointers don't exist in the sandbox already. Is there any reason to change those?

PR #846, opened 2025-02-27
NRL 1320/1321 Update reporting infrastructure

jackleary:


terraform/account-wide-infrastructure/modules/glue/src/pipeline.py
```diff
@@ -47,16 +50,37 @@ def run(self):
             self.logger.error(f"ETL process failed: {e}")
             raise e
 
+    def get_last_run(self):
+        all_runs = self.glue.get_job_runs(JobName=self.job_name)
+        last_runtime = None
```
axelkrastek1-nhs: I would do early exit here instead:
```python
if not all_runs["JobRuns"]:
    return None

for run in all_runs["JobRuns"]
    if run["JobRunState"] == "SUCCEEDED":
        return time.mktime(i["StartedOn"].timetuple())
```

axelkrastek1-nhs (dismissed):

axelkrastek1-nhs (approved):

PR #856, opened 2025-03-07
[NRL-1311] Add script to set pointer supersede info. Add script to put pointers from JSON files

mattdean3-nhs:


scripts/set_pointer_supersede.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+    )
+
+    print(f"Adding superseded info to {pointer_id}...")
+    updated_pointer = DocumentPointer.from_document_reference(doc_ref)
+    table.put_item(
+        Item=updated_pointer.dict(exclude_none=True, exclude={"_from_dynamo"})
+    )
+
+    if delete_superceded:
```
axelkrastek1-nhs: typo: superseded

scripts/put_pointers_from_files.py
```diff
@@ -0,0 +1,43 @@
...
+import json
+
+import fire
+from aws_session_assume import get_boto_session
+
+from nrlf.core.dynamodb.model import DocumentPointer
+from nrlf.core.logger import logger
+from nrlf.producer.fhir.r4.model import DocumentReference
+
+logger.setLevel("ERROR")
```
axelkrastek1-nhs, line 14: Does this logger and the one for the file below get automatically used in the prints?

scripts/set_pointer_supersede.py
```diff
@@ -0,0 +1,96 @@
...
+    supersede_pointer_id: str,
+    delete_superceded: bool = False,
+    env: str = "dev",
+    table_name: str | None = None,
+):
+    session = aws_session_assume.get_boto_session(env)
+    dynamodb = session.resource("dynamodb")
+
+    if not table_name:
+        table_name = f"nhsd-nrlf--{env}-pointers-table"
```
axelkrastek1-nhs, line 41: Do we want to do an extra confirmation when the env is prod?

scripts/set_pointer_supersede.py
```diff
@@ -0,0 +1,96 @@
...
+    except Exception as e:
+        print(f"Could not validate pointer from table. Error: {e}")
+        return
+
+    doc_ref = DocumentReference.model_validate_json(pointer.document)
+
+    if not doc_ref.relatesTo:
+        doc_ref.relatesTo = []
+    else:
+        for relatesTo in doc_ref.relatesTo:
```
axelkrastek1-nhs: What if the `supersede_pointer_id` was typed in wrong? missing an extra character for example, would we want to be extra cautious and print all relatesTo ids for the person to verify before continuing?

scripts/put_pointers_from_files.py
mattdean3-nhs: No - this line stops the debug logging when the script is run.

scripts/set_pointer_supersede.py
mattdean3-nhs: :thinking: not sure, what would we want to achieve with that? A "just in case the env is prod by accident" alert?

scripts/set_pointer_supersede.py
mattdean3-nhs: you mean if the id value isn't correct? like if the ODS code was missing or if it was the wrong length?

axelkrastek1-nhs: Ah my bad, I misinterpreted the code below. This code would stop any pointer that has already been superseding.

The drawback is that this would block the case when a pointer is superseding a superseding pointer which might be a valid case? I guess in that case you would need to modify the script?

scripts/set_pointer_supersede.py
axelkrastek1-nhs: Yeah, precisely that. It might be unnecessary as you need to login into aws prod and also set the `env` variable to prod, so it should be clear at that point you're doing prod.

scripts/set_pointer_supersede.py
mattdean3-nhs: I think it would work fine for a pointer that is superseding a superseding pointer as the supersede info in that case would be on the old pointer (not the current one).

It wouldn't work for the case where the pointer is already superseding another pointer but that's the intention as we don't want pointers with multiple supersedes.

scripts/put_pointers_from_files.py
```diff
@@ -12,10 +13,19 @@
 
 logger.setLevel("ERROR")
 
+SKIP_PROD_WARNING = os.getenv("SKIP_PROD_WARNING", "false")
+
 
 def _put_pointers_from_files(
     *filenames, env: str = "dev", table_name: str | None = None
 ):
+    if env == "dev" and SKIP_PROD_WARNING != "true":
```
axelkrastek1-nhs: Change "dev" for "prod"

axelkrastek1-nhs (approved):

PR #861, opened 2025-03-11
NRL 1342 release reporting infra to int

jackleary:


terraform/infrastructure/locals.tf
```diff
@@ -22,7 +22,7 @@ locals {
   dynamodb_timeout_seconds = "3"
 
   is_sandbox_env = length(regexall("-sandbox-", local.stack_name)) > 0
-  is_dev_env     = var.account_name == "dev"
+  is_dev_env     = var.account_name == "dev" || var.account_name == "int"
```
mattdean3-nhs, line 25: Will this `is_dev_env` flag be removed once we add the reporting bits for all environments?

Only reason I ask is because the `int` env isn't a dev env so this could get confusing in future.

mattdean3-nhs: Confirmed in standup - this is the case and the flag will be removed.

mattdean3-nhs (approved):

PR #866, opened 2025-03-19
NRL-708 fix validation exception errors

eesa456:


layer/nrlf/core/validators.py
```diff
@@ -272,7 +272,7 @@ def _validate_asid(self, asid_references: list):
             return
 
         idx, asid_reference = asid_references[0]
-        asid_value = getattr(asid_reference.identifier, "value", "")
+        asid_value = getattr(asid_reference.identifier, "value") or ""
```
mattdean3-nhs, line 275: this has kinda got me thinking..... given `getattr()` will only set the default if the attribute is not set, does this make a bunch of the other usages of `getattr()` also invalid or ineffective if the attribute value is set to `None`? :thinking:

mattdean3-nhs (approved):
:+1: LGTM for the ASID

General question/comment about our use of getattr in other places but that might be out-of-scope for this specific PR/ticket.

PR #884, opened 2025-04-15
NRL-523 delete response is 200 for no pointer found

eesa456:


layer/nrlf/core/response.py
```diff
@@ -117,6 +117,20 @@ def RESOURCE_DELETED(cls):
             statusCode="200",
         )
 
+    @classmethod
+    def RESOURCE_DOES_NOT_EXIST_DELETE(cls):
```
eesa456, line 121: can alter this however people feel, if we want the diagnostics to say its already deleted or something else thats cool too

axelkrastek1-nhs (approved):

PR #896, opened 2025-04-28
NRL 1411 fix schema issue

jackleary:


terraform/account-wide-infrastructure/modules/glue/iam.tf
```diff
@@ -81,6 +81,16 @@ data "aws_iam_policy_document" "glue_service" {
     effect = "Allow"
   }
 
+  statement {
+    actions = [
+      "cloudwatch:*",
```
mattdean3-nhs: Will glue only need specific "read-only" actions on specific log groups? If that is the case, can we be more specific on the actions/resources that we want to allow glue to use?

Doesn't have to be now thing if that's going to be a blocker, but probably something to make sure we've considered before we go into prod - maybe a ticket if we're going to do it later.

mattdean3-nhs (approved):

PR #903, opened 2025-05-11
NRL 1385 Set up EC2 for automated data refresh

jackleary:


terraform/account-wide-infrastructure/modules/ec2/scripts/user_data.tpl
```diff
@@ -0,0 +1,27 @@
+
+
+<powershell>
+C:\ProgramData\Amazon\EC2-Windows\Launch\Scripts\InitializeInstance.ps1
```
mattdean3-nhs: We should bin this line off maybe? I added it cos I thought it might fix the meta-data issue but it didn't.

terraform/account-wide-infrastructure/modules/ec2/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+    ]
+    effect = "Allow"
+    resources = [
+      "*"
+    ]
+  }
+
+  statement {
+    actions = [
+      "glue:*",
```
mattdean3-nhs, line 83: does the EC2 instance need full access to glue? I thought it'd just use athena to get the data?

terraform/account-wide-infrastructure/modules/ec2/scripts/user_data.tpl
jackleary: Good point I'll get rid

terraform/account-wide-infrastructure/modules/ec2/iam.tf
jackleary: It does, I encountered a perms issue and it listed some Glue action when I was trying to access Athena and the reason is because Athena uses the Glue Data Catalogue to store metadata about the data it queries

mattdean3-nhs (approved):

PR #907, opened 2025-05-22
NRL-1114 for nicip types allow multiple categories

eesa456:


layer/nrlf/core/constants.py
```diff
@@ -180,8 +180,14 @@ def coding_value(self):
...
-    PointerTypes.MRA_UPPER_LIMB_ARTERY.value: Categories.DIAGNOSTIC_STUDIES_REPORT.value,
-    PointerTypes.MRI_AXILLA_BOTH.value: Categories.DIAGNOSTIC_PROCEDURE.value,
+    PointerTypes.MRA_UPPER_LIMB_ARTERY.value: {
+        Categories.DIAGNOSTIC_STUDIES_REPORT.value,
+        Categories.DIAGNOSTIC_PROCEDURE.value,
+    },
+    PointerTypes.MRI_AXILLA_BOTH.value: {
+        Categories.DIAGNOSTIC_PROCEDURE.value,
+        Categories.DIAGNOSTIC_STUDIES_REPORT.value,
+    },
```
katebobyn-nhs, line 190: Open question: is it better to use sets for everything, even those types that map to only one category? Advantage: consistency, simplifies logic as we can use set union without having to check for type. Disadvantage: This branch diverges further from the main branch, and changes can't be ported over as easily. 
Can be picked up later if this work proceeds.

layer/nrlf/core/dynamodb/repository.py
```diff
@@ -33,6 +38,47 @@ def _get_sk_ids_for_type(pointer_type: str) -> tuple:
...
+                category_set.update(cats)
+    return category_set
+
+
+def _build_filter_expressions(
+    pointer_types, categories, expression_names, expression_values
+):
+    """Build DynamoDB filter expressions for pointer_types and categories."""
+    filter_expressions = []
+    if pointer_types:
```
katebobyn-nhs, line 59: For clarity, this will ALWAYS be true, as if the searcher does not specify a a pointer type filter in their search, pointer_types is populated by the list of pointers for which they have permissions. If they have no pointer types permitted, they never get to this point, as it is caught by the decorator and 403 is returned.

layer/nrlf/core/dynamodb/repository.py
```diff
@@ -220,56 +266,36 @@ def search(
...
+            filter_expressions = _build_filter_expressions(
+                pointer_types, categories, expression_names, expression_values
+            )
+        elif pointer_types and not categories:
+            # Get all categories for these pointer_types
+            all_categories = list(_get_categories_for_pointer_types(pointer_types))
+            filter_expressions = _build_filter_expressions(
+                pointer_types, all_categories, expression_names, expression_values
+            )
+        elif categories and not pointer_types:
```
katebobyn-nhs, line 294: Similar to the above comment, pointer_types will always be populated (whether by a chosen filter or by the searcher's list of permitted pointers) so this will never get reached.

layer/nrlf/core/dynamodb/repository.py
```diff
@@ -220,56 +266,36 @@ def search(
...
-            expression_values.update(types_filter_values)
+        # Determine which filters to apply
+        if pointer_types and categories:
+            # Use both pointer_types and categories as filters
+            filter_expressions = _build_filter_expressions(
+                pointer_types, categories, expression_names, expression_values
+            )
+        elif pointer_types and not categories:
+            # Get all categories for these pointer_types
+            all_categories = list(_get_categories_for_pointer_types(pointer_types))
```
katebobyn-nhs, line 290: Open question: Is there value to looking up all possible categories and adding filters for them, if type(s) already specified?

katebobyn-nhs (approved):
Following discussion in dev-cop, putting a pin in this since it's unclear whether this will be needed for the POC. Confirmed this works, there are some open questions about whether it's the best approach for the use case but these cannot really be answered until the use case is better defined -- we are unsure if XCA team will be reorganising the type and category codes so that modality, laterality and body site are searchable via context filters (thus the NICIP codes would not be needed and we can keep the 1-to-may category:type mapping).

PR #908, opened 2025-05-22
NRL-1205 dont ignore category if type is given for search

eesa456:


api/consumer/searchDocumentReference/tests/test_search_document_reference_consumer.py
```diff
@@ -247,6 +247,60 @@ def test_search_document_reference_happy_path_with_category(
...
+    ).get("display")
+    doc_ref2.category[0].coding[0].code = Categories.OBSERVATIONS.coding_value()
+    doc_ref2.category[0].coding[0].display = CATEGORY_ATTRIBUTES.get(
+        Categories.OBSERVATIONS.value
+    ).get("display")
+    repository.create(DocumentPointer.from_document_reference(doc_ref2))
+
+    event = create_test_api_gateway_event(
+        headers=create_headers(),
+        query_string_parameters={
```
katebobyn-nhs, line 274: If I'm reading this right, the test is 
1. Creating a pointer of type: MHCP, cat: Care Plan
2. Creating a pointer of type: NEWS2, cat: Obs
3. Searching for pointers of type: MHCP, cat: Care Plan
It then expects the one result (pointer 1) since pointer 2 would be filtered out by both criteria. 

I think to prove both filters are being applied there needs to be a test scenario where, e.g., you search for type: MHCP, cat: Obs
Then the expected result is an empty searchset bundle because 1 is filtered out on the basis of failure to match category, and 2 is filtered out on the basis of failure to match type.

layer/nrlf/core/dynamodb/repository.py
```diff
@@ -226,41 +226,16 @@ def search(
...
+            categories=categories,
         )
 
         key_conditions = ["patient_key = :patient_key"]
         filter_expressions = []
         expression_names = {}
         expression_values = {":patient_key": f"P#{nhs_number}"}
 
-        if len(pointer_types) == 1:
-            # Optimisation for single pointer type
```
katebobyn-nhs, line 237: I think this optimisation might still make sense in this implementation
but happy to discuss

katebobyn-nhs:
Can we add feature tests?

layer/nrlf/core/dynamodb/repository.py
eesa456: it can work, my worry was that it would cause other categories to be ignored in the search due to the begins with condition. This isnt really an issue if we only have the one category to type relationship but if we have more than one category per type this wouldnt work (something i found on the NIR work on the other PR) guess I was future proofing but honestly I have no issues adding it back so I'll do that

eesa456: ALSO the search function here believes that we can have multiple types passed in a list but actually when we do something like type=type1,type2 in the search params this doesnt work... because of validate_type in validators which we do on all search lambdas, it checks that the type param is in the orgs pointer types -> this only works if one type is given the second you add more it doesnt work because its treating it a string type and "type1,type2" isnt going to be in the list even if type1 and type2 both are.

So the only time we hit the code for pointer_types > 1 is when we dont have the type filter in the params. I guess is that on purpose and we dont want them to be searching for more than one kind of type at once ?? This is more so a question for me to understand how it should work FHIR wise.

api/consumer/searchDocumentReference/tests/test_search_document_reference_consumer.py
eesa456: Yeah I'll add that I'll also add searching with multiple categories and the one type to show that the optimisation for the single type is still used and that it wont get other categories even if they're provided because it isnt for that type

eesa456:
> Can we add feature tests?

There are already some feature tests I wrote back when I did the category filter work but I can do an additional scenario that is basically the unit test you mentioned above but an integration test

layer/nrlf/core/dynamodb/repository.py
axelkrastek1-nhs: Implementation was put back in.

mattdean3-nhs (approved):
LGTM :+1:

PR #918, opened 2025-05-28
[NRL-1470] Add retries to test api client for 502 errors

mattdean3-nhs:


tests/utilities/api_clients.py
```diff
@@ -34,6 +36,35 @@ def add_pointer_type(self, pointer_type: PointerTypes):
...
+            for attempt in range(2):
+                response = func(*args, **kwargs)
+                if not response.status_code or response.status_code not in status_codes:
+                    return response
+                attempt_responses.append(response)
+                print(  # noqa: T201
+                    f"Retrying due to {response.status_code} error in attempt {attempt + 1}..."
+                )
+
+            print(  # noqa: T201
```
katebobyn-nhs: Discussed IRL but would be nice to get these on std out so that we still get the 'smoke' if there are intermittent failures, just that it doesn't fail the build

mattdean3-nhs: Yep - good shout. I've switched the `print()` calls to use `logging` and you can now see the output in the test results:

image

katebobyn-nhs (approved):

PR #932, opened 2025-06-13
NRL-497 remove count endpoint and add _summary filter to search endpoint

eesa456:
…ints


api/consumer/searchDocumentReference/search_document_reference.py
```diff
@@ -102,6 +105,28 @@ def handler(
...
+        bundle = {
+            "resourceType": "Bundle",
+            "type": "searchset",
+            "link": [{"relation": "self", "url": self_link}],
+            "total": 0,
+        }
+        logger.log(LogReference.CONSEARCH006)
+        total = sum(
+            1
+            for _ in repository.search(
```
mattdean3-nhs: Downside to using `repository.search()` over `repository.count_by_nhs_number()` is that it returns all items. The latter uses `SELECT: COUNT` for the DynamoDB query, meaning no items are actually returned from the query so it's faster, lighter and lower cost.

However, we don't current expect any NHS numbers to have more than a few pointers, and most will have none so this could be a little premature. I'll have a think/chat about this, we can decide on it as a team.

tests/smoke/scenarios/consumer_count_search_read.py
```diff
@@ -46,13 +45,6 @@ def test_consumer_count_search_read(
     patient_id = test_data["patient_nhs_number"]
     test_pointers = test_data["pointers"]
 
-    # Count
-    count_response = consumer_client.count(
```
mattdean3-nhs, line 50: should we add an extra call to search with the new `_summary` param to mimic the count in the smoke tests..... :thinking:

tests/performance/consumer/client.js
```diff
@@ -34,22 +34,6 @@ function checkResponse(res) {
   }
 }
 
-export function countDocumentReference() {
```
mattdean3-nhs: I think we should keep this function but refactor it to use the new search with _summary so we can keep records of the perf of the count search.

api/producer/searchDocumentReference/search_document_reference.py
```diff
@@ -86,6 +86,24 @@ def handler(
         categories=params.category.root.split(",") if params.category else [],
     )
 
+    if params.field_summary and params.field_summary.root == "count":
```
mattdean3-nhs: love this - great idea adding it for producer API too :muscle:

tests/smoke/scenarios/consumer_count_search_read.py
mattdean3-nhs: Decided not to do this - might need to change the filename to `consumer_search_read.py` as there's no longer a count in there.

mattdean3-nhs (dismissed):
:+1: LGTM - need to resolve the conflicts

Only remaining comment is to rename the smoke-test file to remove `count` as it doesn't do a count anymore.

mattdean3-nhs (dismissed):

mattdean3-nhs (approved):
As before - approving branch updates from develop

PR #937, opened 2025-06-18
NRL-1386 Add infra and config for prod dashboards

mattdean3-nhs:
Main points to call out:
- Resources for reporting, like athena and glue, are now provisioned once for each AWS account
- TF var added to allow active reporting services (like glue and EC2) to be enabled/disabled account-wide
- Reporting source bucket is no longer versioned - to fix messy S3 lifecycle delete-marker issues
- Firehose streams are no longer created in ephemeral envs
- Fixed state issues with API gateway 403 responses


terraform/account-wide-infrastructure/test/athena.tf
```diff
@@ -1,5 +1,35 @@
+/*module "qa-athena" {
```
mattdean3-nhs: TODO - Once we've decided that single Athena per AWS account is what we want, we should delete these before the PR is merged.

terraform/account-wide-infrastructure/test/glue.tf
```diff
@@ -1,5 +1,30 @@
+/*module "qa-glue" {
```
mattdean3-nhs: TODO - Once we've decided that single Glue per AWS account is what we want, we should delete these before the PR is merged.

terraform/account-wide-infrastructure/dev/vars.tf
```diff
@@ -14,6 +14,12 @@ variable "devsandbox_api_domain_name" {
   default     = "dev-sandbox.api.record-locator.dev.national.nhs.uk"
 }
 
+variable "enable_reporting" {
+  type        = bool
+  description = "Enable account-wide reporting services in the dev account"
```
mattdean3-nhs: Re-word this to make it a bit clearer that it's not resources, it's processing that is enabled/disabled

terraform/account-wide-infrastructure/modules/powerbi-gw-ec2/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "instance_id" {
+  value = aws_instance.powerbi_gw.id
```
mattdean3-nhs: Check we still need this

terraform/infrastructure/vars.tf
```diff
@@ -48,3 +55,9 @@ variable "log_retention_period" {
   default = 90
   type    = number
 }
+
+variable "enable_reporting" {
```
mattdean3-nhs, line 59: Unused - we use the shared-resources flag instead , so can be removed

axelkrastek1-nhs (dismissed):

axelkrastek1-nhs (approved):

PR #938, opened 2025-06-19
NRL-512 Initial implementation for rejecting HEAD requests

axelkrastek1-nhs:


terraform/infrastructure/modules/api_gateway/method_responses.tf
```diff
@@ -0,0 +1,68 @@
...
+}
+
+resource "aws_api_gateway_integration" "head_integration" {
+  for_each = var.endpoint_allowed_methods
+
+  rest_api_id = aws_api_gateway_rest_api.api_gateway_rest_api.id
+  resource_id = data.aws_api_gateway_resource.resource[each.key].id
+  http_method = aws_api_gateway_method.head_method[each.key].http_method
+  type        = "MOCK"
+  # passthrough_behavior = "WHEN_NO_TEMPLATES"
```
katebobyn-nhs: Guessing we can remove this as you've gone with the WHEN_NO_MATCH option. 
As far as I understand, the only differences in behaviours between the two would be when the content type was anything other than json/FHIR combos (because those are the ones that have templates defined below): 
NO_TEMPLATES would return a 415 (because there are templates, but none that match the content type)
whereas
NO_MATCH will pass the requests through to the API for handling (because it didn't match the templates)
Either seems ok, and as far as I understand, if no content type is specified APIG defaults to application/json so the behaviour would be the same in most cases.

terraform/infrastructure/modules/api_gateway/method_responses.tf
```diff
@@ -0,0 +1,68 @@
+# Define a reusable map of allowed methods for each path
```
katebobyn-nhs: Just noting this is a newly created file whereas there are other API Gateway method responses in terraform/infrastructure/modules/api_gateway/api_gateway.tf 
Is this one meant just for the HEAD responses (and if so can we name it so nobody is looking for other method responses here)? Or should we move the other ones here too?

terraform/infrastructure/api_gateway.tf
```diff
@@ -35,7 +39,10 @@ module "producer__gateway" {
     method_deleteDocumentReference     = "arn:aws:apigateway:eu-west-2:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-2:${local.aws_account_id}:function:${substr("${local.prefix}--api--producer--deleteDocumentReference", 0, 64)}/invocations"
     method_status                      = "arn:aws:apigateway:eu-west-2:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-2:${local.aws_account_id}:function:${substr("${local.prefix}--api--producer--status", 0, 64)}/invocations"
   }
-
+  endpoint_allowed_methods = {
+    "/DocumentReference"      = "GET,POST",
+    "/DocumentReference/{id}" = "GET,PUT,DELETE"
```
katebobyn-nhs, line 44: I completely missed DELETE in the ACs, thank you for being thorough. I've updated it now.

katebobyn-nhs (commented):
Looks good. Made some comments/notes. We can discuss in the morning but nothing blocking. 

Another minor suggestion would be to update `terraform/infrastructure/consumer.tftpl` and `terraform/infrastructure/producer.tftpl` to note this deviation from the default behaviour (which would be to actually handle HEAD reqs wherever we support GET). 
I think it would suffice to add a sentence in the DocumentReference 'documentation' string so it would now read, 
"Additional business rules apply to constrain patient, organisation and type/category content. **Note that HEAD requests are not supported.**"

terraform/infrastructure/modules/api_gateway/method_responses.tf
axelkrastek1-nhs: Good point, I can change the name

terraform/infrastructure/modules/api_gateway/method_responses.tf
axelkrastek1-nhs: Good catch, sorry, this was a leftover when I was testing things. I'll remove it

axelkrastek1-nhs: After discussions, possibly best to leave WHEN_NO_TEMPLATES instead and add a feature test when contentType is not defined, which should get us a 415

axelkrastek1-nhs: Add an integration test where content-type is not set.

axelkrastek1-nhs: And an integration test when the content-type is not part of the below list

eesa456 (dismissed):

mattdean3-nhs (approved):
Re-approving due to branch updates - no code changes since last approval.

PR #939, opened 2025-06-20
[NRL-1379] Add TF config for lambda error notifications in test and prod

mattdean3-nhs:


terraform/account-wide-infrastructure/prod/locals.tf
```diff
@@ -3,4 +3,6 @@ locals {
   project     = "nhsd-nrlf"
   environment = terraform.workspace
   prefix      = "${local.project}--${local.environment}"
+
+  notification_emails = nonsensitive(toset(tolist(jsondecode(data.aws_secretsmanager_secret_version.emails.secret_string))))
```
axelkrastek1-nhs: I see this is the same as the dev account. So I understand for consistency it should be like this but it feels like this shouldn't be marked as nonsensitive?

axelkrastek1-nhs (dismissed):

terraform/account-wide-infrastructure/prod/locals.tf
mattdean3-nhs: Yep good spot - have fixed so the value is sensitive all the time.

terraform/account-wide-infrastructure/dev/secrets.tf
```diff
@@ -7,10 +7,6 @@ resource "aws_secretsmanager_secret" "backup_destination_parameters" {
   description = "Parameters used to configure the backup destination"
 }
 
-resource "aws_secretsmanager_secret" "notification_email_addresses" {
-  name = "${local.prefix}-dev-notification-email-addresses"
-}
-
```
axelkrastek1-nhs, line 13: Good catch 🙌

axelkrastek1-nhs (approved):

PR #947, opened 2025-06-24
APM-000-AMP-3654-removed interactive-product-backlog

nkinsler1:


mattdean3-nhs:
We've removed these sections from the NRLF swagger - to affect the public documentation, please see repos `nrl-consumer-api` and `nrl-producer-api`

PR #967, opened 2025-07-07
NRL-753 fix capability statement endpoint deployment

eesa456:


mattdean3-nhs (dismissed):
:+1: happy with this - it resolves the dependency issues which we think will resolve deployment issues with APIGW, including the cap statement stuff

mattdean3-nhs (approved):

PR #1017, opened 2025-08-22
NRL-1541 add authorised organisations & actors catalogue

DruckerDeming:
NRL-1541-authorised-organisations-&-actors-catalogue


DruckerDeming:
Created another branch, taking a different approach

PR #1026, opened 2025-09-02
NRL-1587 fix cron job invalid expression

hamzakhan24-nhs:
pair implementation with Sandy F (due to no GitHub access currently)


mattdean3-nhs (approved):
lgtm 👍

PR #1043, opened 2025-09-22
[NRL-1606] Add report to get masterids for pointers

mattdean3-nhs:


reports/get_pointer_masterids_for_custodian.py
```diff
@@ -0,0 +1,97 @@
...
+            master_id = item.get("master_identifier", {}).get("S", "no-master-id")
+            custodian = item.get("custodian", {}).get("S", "no-custodian")
+
+            pointers_info.append(
+                {
+                    "nrl-id": pointer_id,
+                    "pointer-type": pointer_type,
+                    "master_identifier": master_id,
+                    "custodian": custodian,
+                    "patient_id": item.get("nhs_number", {}).get("S", "no-patient-id"),
```
axelkrastek1-nhs: `patient_id` is always included here, you need to check `INCLUDE_PATIENT_IDS`

reports/get_pointer_masterids_for_custodian.py
```diff
@@ -0,0 +1,97 @@
...
+        for item in page["Items"]:
+            pointer_id = item.get("id", {}).get("S", "no-id")
+            pointer_type = item.get("type_id", {}).get("S", "no-type")
+            master_id = item.get("master_identifier", {}).get("S", "no-master-id")
+            custodian = item.get("custodian", {}).get("S", "no-custodian")
+
+            pointers_info.append(
+                {
+                    "nrl-id": pointer_id,
+                    "pointer-type": pointer_type,
```
axelkrastek1-nhs, line 64: All pointer types are included but the ticket only talks about EOL pointer type, maybe add an optional filter for types in the FilterExpression?

reports/get_pointer_masterids_for_custodian.py
```diff
@@ -0,0 +1,97 @@
...
+
+    print(  # noqa
+        f"Getting masterids for custodians {custodian_list} in table {table_name}...."
+    )
+
+    required_attributes = (
+        ["id", "type_id", "master_identifier", "custodian"]
+        if not INCLUDE_PATIENT_IDS
+        else ["id", "type_id", "master_identifier", "custodian", "nhs_number"]
+    )
```
axelkrastek1-nhs: nitpick:
```
required_attributes = ["id", "type_id", "master_identifier", "custodian"]
if INCLUDE_PATIENT_IDS:
    required_attributes.append("nhs_number")
```

reports/get_pointer_masterids_for_custodian.py
mattdean3-nhs: I think that's ok - the `.get(type, default_value)` will return a default value of `no-patient-id` if `INCLUDE_PATIENT_IDS` is not set

reports/get_pointer_masterids_for_custodian.py
mattdean3-nhs: yeah, was thinking that originally too but thought it'd make the report simpler to just get all pointers and filter on the client side before sending the data.

What do you reckon?

reports/get_pointer_masterids_for_custodian.py
mattdean3-nhs: I personally prefer my pythonic ternary version but I can see what you mean, bit of a dupe lists, maybe less readable for non-python peeps..... I'll change it

reports/get_pointer_masterids_for_custodian.py
mattdean3-nhs: Have changed it to return "not-included" if the flag is not set

reports/get_pointer_masterids_for_custodian.py
axelkrastek1-nhs: Filtering afterwards is also ok.

axelkrastek1-nhs (approved):

PR #1059, opened 2025-10-07
NRL-1658 Fix capability statement deployment

mattdean3-nhs:


api/consumer/swagger.yaml
```diff
@@ -382,6 +382,30 @@ paths:
...
+        type: MOCK
+        requestTemplates:
+          application/json: '{ "statusCode": 200 }'
+        responses:
+          default:
+            statusCode: "200"
+            responseTemplates:
+              # prettier-ignore
+              application/json: '${capability_statement_json}'
+        passthroughBehavior: when_no_match
```
axelkrastek1-nhs: Are there feature tests for the capability statement? Is there any where the content-type doesn't match?

sandyforresternhs (commented):
Do we think the existing smoke test is still valid, and if so, would it be worth removing the skip to see if it's had the desired effect?

api/consumer/swagger.yaml
mattdean3-nhs: I don't believe we do - but that's a good spot as it looks like the current capability statement integration does not define this parameter. I've done some testing and the results are:

- parameter not defined - results in 500 error with operation outcome response body
- parameter set to when_no_match - results in 500 error with generic "Internal server error" response

While neither are ideal, I'm going to try and maintain the existing behaviour and just focus on fixing the deployment issues.

Also, no plans to add feature tests for this but happy to be convinced otherwise....

sandyforresternhs (approved):

PR #1070, opened 2025-10-15
[NRL-1700] Add rotation tools to truststore.sh

mattdean3-nhs:


scripts/truststore.sh
```diff
@@ -14,8 +22,14 @@ function _truststore_help() {
     echo "  build-cert <name> <ca> <fqdn>   - Build a single client cert + private key"
     echo "  build-all                       - Build the standard trust store certs"
     echo "  pull-ca <ca>                    - Pull the certificate authority"
+    echo "  pull-ca-key <ca>                - Pull the certificate authority private key"
     echo "  pull-client <env>               - pull the files needed for a client connection"
     echo "  pull-server <env>               - pull the files needed for a server connection"
+    echo "  rotate-ca <env>                 - rotate the certificate authority, archiving the previous one"
+    echo "  rotate-client <env>             - rotate the client certificate, archiving the previous one"
```
sandyforresternhs: This doesn't match up with line 352 - should this be rotate-cert?

scripts/truststore.sh
```diff
@@ -14,8 +22,14 @@ function _truststore_help() {
...
+    echo "  disable-archived-ca <env>       - disable an archived certificate authority"
+    echo "  restore-archived-ca <env>       - restore an archived certificate authority"
+    echo "  restore-archived-client <env>   - restore an archived client certificate"
```
sandyforresternhs: This doesn't match up with line 355 - should this be restore-archived-cert?

scripts/truststore.sh
```diff
@@ -14,8 +22,14 @@ function _truststore_help() {
...
     echo
```
sandyforresternhs, line 35: It's not something you've omitted but there's no mention of pull-all or push-all in the help function - maybe worth adding?

scripts/truststore.sh
```diff
@@ -31,119 +45,293 @@ EOF"
...
-    _truststore_build_ca "prod" "record-locator.national.nhs.uk"
-    _truststore_build_ca "int"  "record-locator.int.national.nhs.uk"
-    _truststore_build_ca "ref"  "record-locator.ref.national.nhs.uk"
-    _truststore_build_ca "qa"   "qa.record-locator.national.nhs.uk"
-    _truststore_build_ca "dev"  "record-locator.dev.national.nhs.uk"
+    _truststore_build_ca "prod" "record-locator.national.nhs.uk_CA2"
+    _truststore_build_ca "int"  "record-locator.int.national.nhs.uk_CA2"
+    _truststore_build_ca "ref"  "record-locator.ref.national.nhs.uk_CA2"
+    _truststore_build_ca "qa"   "qa.record-locator.national.nhs.uk_CA2"
+    _truststore_build_ca "dev"  "record-locator.dev.national.nhs.uk_CA2"
```
sandyforresternhs, line 236: Do we still need the CA2?

sandyforresternhs (commented):
Looking good, just a couple of q's

scripts/truststore.sh
mattdean3-nhs: it should - will fix :+1:

scripts/truststore.sh
mattdean3-nhs: yep - will add

scripts/truststore.sh
mattdean3-nhs: it should - will fix :+1:

scripts/truststore.sh
mattdean3-nhs: we do as that's the current name of the CA we're using

sandyforresternhs (approved):

PR #1071, opened 2025-10-15
[NRL-1700] Log info about which clientcert is being used for each req…

mattdean3-nhs:
…uest


sandyforresternhs (dismissed):

sandyforresternhs:
Lgtm.  I've given this a tick despite the PR deploy check having failed because I believe that's due to the certificate changes needing to be merged first in order for the check to pass.  Suggest we rerun the check once the cert changes have merged.

sandyforresternhs (approved):

PR #1072, opened 2025-10-17
NRL-1631 Add warning log for unexpected multiple pointers

mattdean3-nhs:


api/producer/upsertDocumentReference/upsert_document_reference.py
```diff
@@ -262,6 +263,24 @@ def handler(
...
+            nhs_number, [pointer_type]
+        )
+
+        if existing_pointers_count > 0:
+            logger.log(
+                LogReference.PROUPSERT012,
+                new_pointer_id=core_model.id,
+                pointer_type=pointer_type,
+                nhs_number=nhs_number,
+                existing_pointers_count=existing_pointers_count,
```
sandyforresternhs, line 281: Just wondering about this - this will log the 'before' count - is that what we want, or would the 'after' count be more useful?

Also, if it's an upsert, would there always be at least one existing pointer?

api/producer/createDocumentReference/create_document_reference.py
```diff
@@ -255,6 +256,25 @@ def handler(
...
+            nhs_number, [pointer_type]
+        )
+
+        if existing_pointers_count > 0:
+            logger.log(
+                LogReference.PROCREATE012,
+                new_pointer_master_id=pointer_master_identifier,
+                pointer_type=pointer_type,
+                nhs_number=nhs_number,
+                existing_pointers_count=existing_pointers_count,
```
sandyforresternhs, line 274: Just wondering about this - this will log the 'before' count - is that what we want, or would the 'after' count be more useful?

api/producer/upsertDocumentReference/upsert_document_reference.py
mattdean3-nhs: The "after" count might be useful if we want to use the number for counts/reports.... I think a +1 to the existing_pointers_count would do the trick for that, is that what you were thinking?

Rather confusingly, upsert (in this case) doesn't mean "update or insert". Mulitples-wise, it's pretty much the same as create.

sandyforresternhs (approved):
Code looks good - don't forget you need to update the docs!

https://nhsd-confluence.digital.nhs.uk/spaces/CLP/pages/874679033/NRL+Futures+v3.0+Log+Points

PR #1086, opened 2025-11-04
NRL-1664 New NFT/Performance/Load terraform environment

axelkrastek1-nhs:


terraform/infrastructure/locals.tf
```diff
@@ -32,15 +32,15 @@ locals {
 
   reporting_bucket_arn = data.aws_s3_bucket.source-data-bucket.arn
   reporting_kms_arn    = data.aws_kms_key.glue.arn
-  firehose_lambda_subscriptions = var.use_shared_resources ? (
+  firehose_lambda_subscriptions = var.use_shared_resources && !var.disable_firehose_lambda_subscriptions ? (
```
mattdean3-nhs, line 35: is the `enable_reporting` flag not working here? I think that was meant to be the flag that disables the firehose subscriptions

mattdean3-nhs (commented):
all looks good to me 👍 just one question about the new reporting flag

tests/utilities/get_access_token.py
```diff
@@ -47,6 +48,7 @@
     "ref": "ref",
     "int": "int",
     "int-sandbox": "sandbox",
+    "perftest": "internal-qa",  # TODO: Hijacking qa to not create a new APIGEE proxy environment. Once we have proxygen, rethink this
```
mattdean3-nhs: Also, noticed that this # TODO causes a sonar cloud warning, maybe worth just leaving the comment and removing the TODO bit?

terraform/infrastructure/locals.tf
axelkrastek1-nhs: Unfortunately no. For shared environments, the default when `enable_reporting = false` is to enable splunk.

mattdean3-nhs: ah ok yeah, that makes sense as reporting is different to splunk :cry: Let's keep this for now and can revise if we have issues in future.

mattdean3-nhs (dismissed):
LGTM

Happy to re-approve if you fix that sonarcloud comment issue but not super worried about that either way.

tests/utilities/get_access_token.py
axelkrastek1-nhs: I did it as a TODO to not forget about it in the future, but I think what I'll do is raise a ticket and put the reference here instead 👍

atrace (approved):

PR #1096, opened 2025-11-17
NRL-1798 Add delete pointers script

sandyforresternhs:


scripts/delete_pointers_by_id.py
```diff
@@ -0,0 +1,317 @@
...
+
+    # Proceed with deletion using BatchWriteItem
+    pointers_deleted, failed_deletes = _batch_delete_pointers(
+        table_name, existing_pointers
+    )
+
+    end_time = datetime.now(tz=timezone.utc)
+
+    print(" Done")
+    return {
```
mattdean3-nhs: I've just done a few tests of the script, when there are many pointers (hundereds/thousands), the output from the command might be difficult to parse or use.

In some of our other scripts, we sometimes output a `.json` file containing the larger data outputs and just display a summary on the CLI output, that could be an option here. What do you think?

mattdean3-nhs (commented):
Code and units tests look good to me 💪

One question around command output

There's just one sonarcloud warning that needs looking at: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=1096&id=NHSDigital_NRLF&open=AZqSqq1_aAIFdLwvl4vZ

scripts/delete_pointers_by_id.py
sandyforresternhs: Great call, I'll add that 👍

mattdean3-nhs (approved):
LGTM 👍 nice work!

scripts/delete_pointers_by_id.py
sandyforresternhs: Output file added

PR #1100, opened 2025-11-27
NRL-1595 Experimenting with cert envs needed to deploy account changes

anjalitrace2-nhs:
Turns out we don't need the sandbox truststore certs in order to run this pipeline (and the pipeline is unable to retrieve them!)

Fixes the failing workflows on develop. Passing run of the latest commit [here](https://github.com/NHSDigital/NRLF/actions/runs/19711501750)


mattdean3-nhs (approved):
LGTM ✔️

PR #1111, opened 2025-12-15
NRL-1860 Tweak prep scripts to support NFT runs

mattdean3-nhs:


scripts/seed_nft_tables.py
```diff
@@ -155,25 +159,41 @@ def _populate_seed_table(
...
         if batch_counter + pointers_for_px > 25 or px_counter == px_with_pointers:
-            resource.batch_write_item(RequestItems={table_name: batch_upsert_items})
+            response = resource.batch_write_item(
+                RequestItems={table_name: batch_upsert_items}
+            )
+
+            if response.get("UnprocessedItems"):
+                unprocessed_count += len(
+                    response.get("UnprocessedItems").get(table_name, [])
+                )
```
anjalitrace2-nhs, line 195: ~~Is there any value in tracking not just the number of unprocessed items but also which items were unprocessed? e.g. a list of pointers we failed to create.~~

~~If we fail to insert a couple items we no longer have sequential data - which the perf tests themselves rely on - but also I guess we can't do much with that info~~

I have talked myself out of this - ignore!

scripts/seed_nft_tables.py
```diff
@@ -189,54 +209,92 @@ def _populate_seed_table(
...
+#    """
+#
+#    extra_per_hundred = int(
+#        (pointers_per_px - 1.0) * 100
+#    )  # no patients can have zero pointers
+#    counts = {}
+#    counts["3"] = extra_per_hundred // 10
+#    counts["2"] = extra_per_hundred - 2 * counts["3"]
+#    counts["1"] = 100 - counts[2] - counts[3]
+#    return _set_up_cyclical_iterator(counts)
```
anjalitrace2-nhs: super minor: do we need this or can we bin it?

anjalitrace2-nhs (dismissed):
One minor non-blocking comment, looks good to me! ✅

anjalitrace2-nhs (approved):

PR #1114, opened 2025-12-29
NRL-1811 do not log multiple pointer warning if test patient

katebobyn-nhs:


sandyforresternhs (commented):
Changes look good 👍 .  We've got tests that assert logs are created when multiple pointers are detected - would it be worth adding an assertion to show it doesn't log in relation to test patients?

sandyforresternhs (approved):

PR #1119, opened 2026-01-13
NRL-1884 seed script generates perf test input files

anjalitrace2-nhs:
- seed script will now generate the needed input files for perf tests on the fly and upload these to S3 (along with some metadata about when the file was generated e.g. commit hash). This enables the seed script to be run from within our new bastion & saves us the previous computationally expensive generation which scanned the entire pointer table 3 times
- repurpose generate_pointer_table_extract function to be useful for refreshing these input file based on current table state. Ran this for baseline table.
- perf tests prepare step will grab the input files from s3 and regenerate distribution info before a test run (cheap to do)
- get rid of unused input files & steps
- fix a bug in the seed script where the final batch of writes were never written (`patient_counter <= patients_with_pointers` rather than `<`)
- renamed px > patient(s) for my brain
- updated readme instructions to reflect updated make commands

Tested seed stuff works using a tiny test table anjali-test-2 by seeding ~20 pointers and checking files generated. Ran prepare step and ran (most - all successes 30 mins in) perf tests and it was all able to use the newly generated input files & download them from s3


Makefile
```diff
@@ -249,31 +253,39 @@ generate-models: check-warn ## Generate Pydantic Models
...
 	@echo "Generating pointer table extract with PERFTEST_TABLE_NAME=$(PERFTEST_TABLE_NAME) and DIST_PATH=$(DIST_PATH)"
-	mkdir -p $(DIST_PATH)
-	PYTHONPATH=. poetry run python tests/performance/perftest_environment.py generate_pointer_table_extract --output_dir="$(DIST_PATH)"
-
-perftest-prepare: perftest-prep-generate-producer-data perftest-prep-extract-consumer-data perftest-prep-generate-pointer-table-extract
-	@echo "Prepared performance tests with PERFTEST_TABLE_NAME=$(PERFTEST_TABLE_NAME) and DIST_PATH=$(DIST_PATH)"
+	rm -rf "${DIST_PATH}/nft"
+	mkdir -p "${DIST_PATH}/nft"
+	PYTHONPATH=. poetry run python tests/performance/perftest_environment.py generate_pointer_table_extract --output_dir="${DIST_PATH}/nft"
+	./scripts/get-current-info.sh > "${DIST_PATH}/nft/info.json"
```
mattdean3-nhs, line 289: will this `info.json` file get overwritten by the file in the zip when it's extracted?

tests/performance/producer/client_perftest.js
```diff
@@ -7,8 +7,8 @@ import { createRecord } from "../setup.js";
 import exec from "k6/execution";
 
 const csvPath = __ENV.DIST_PATH
-  ? `../../../${__ENV.DIST_PATH}/producer_reference_data.csv`
-  : "../producer_reference_data.csv";
+  ? `../../../${__ENV.DIST_PATH}/seed-pointers-extract.csv`
```
mattdean3-nhs: should this be the `.csv` file in the `./dist/nft` directory? the one that comes out of the .zip file? (same in `consumer/client_perftest.js` too)

mattdean3-nhs (commented):
Love this 😍 it works a treat.

Only issue I hit while running was that the seed-pointers-extract.csv file wasn't there (see comment for client_perftest.js).

tests/performance/producer/client_perftest.js
anjalitrace2-nhs: this is that same file - it's [copied out of the unzipped directory](https://github.com/NHSDigital/NRLF/blob/16bda8d3603f0d7adc2374b8e36147c753182cfd/Makefile#L273) in the prepare step (mostly to remove the table name from the file name).

This is a bit confusing though! I'll skip the copy step and use the file inside the ./dist/nft instead 👍

Makefile
anjalitrace2-nhs: yes, just before the zip is extracted we [clear the whole ./dist/nft folder](https://github.com/NHSDigital/NRLF/blob/NRL-1884-seed-script-generates-input-files/Makefile#L269) too, so w'll have a fresh version of every file in there

tests/performance/producer/client_perftest.js
anjalitrace2-nhs: done ✅

mattdean3-nhs (approved):

PR #1122, opened 2026-01-22
NRL-1875 Make output better for tests & seed script

anjalitrace2-nhs:
~~Marked as draft because I'm using this branch to alter seed runs on the bastion and would be nicer to wait until the branch this merges into ([feature/SAFO6-NRL-1841-performance-tests-public-mode](https://github.com/NHSDigital/NRLF/tree/feature/SAFO6-NRL-1841-performance-tests-public-mode)) is in develop for less mess!~~ Tidied up & ready for review now

- Make perf tests give a more detailed output report including a breakdown per scenario
- The csv generated by the seed script now does not include any unprocessed items that failed to be upserted to dynamo
    - Still not an accurate representation of pointer count increase - items which were updated rather than created will count as processed and therefore be included in this csv too
- When regenerating the input csv, cap it by default at 1.5m rows. As long as there are more rows than test iterations, this is sufficient as input for the perf tests. Generally we have 9k-14k iterations. Trying to load a 15m row csv into the test VUs was very slow and ate up >350% of my machine's CPU before the tests had even begun!
- Point perftest env at 15m table for 15m testing
    - Updated the readme's instructions on how to do this
- Ramped the TPS down to 1 per scenario. Scenarios run in parallel so this gives roughly producer = 5 TPS, consumer = 7 TPS.
- Redid the baseline testing with these new TPS so that we could accurately compare to higher volumes


sandyforresternhs (approved):
LGTM :)

PR #1123, opened 2026-01-23
NRL-1417 add SBOM generation step to nightly build

anjalitrace2-nhs:
Uses the shared trivy action to generate a new SBOM every night from develop.

The generated SBOM is stored in github as an artifact against the action run that generated it. We can also run the workflow ad-hoc if we ever cannot wait until the next scheduled run to refresh it. See our first SBOM here: https://github.com/NHSDigital/NRLF/actions/runs/21292962386 (scroll down)

Linked PR for the shared trivy action to allow an SBOM to be generated from a git repo rather than only a docker image https://github.com/nhs-england-tools/trivy-action/pull/10 (merged & released in v1.4.0)

### DIFF
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -38,11 +38,6 @@ jobs:
       - name: Run Unit Tests
         run: make test
 
-      - name: Generate SBOM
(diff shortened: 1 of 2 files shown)


mattdean3-nhs (approved):
LGTM 👍

PR #1124, opened 2026-01-26
NRL-0000 Upgrade EC2 Storage

jackleary:
Upgrade both storage type and space for refresh


terraform/account-wide-infrastructure/dev/vars.tf
```diff
@@ -67,3 +67,15 @@ variable "use_powerbi_gw_custom_ami" {
   description = "Use custom image for PowerBI GW instance"
   default     = true
 }
+
+variable "root_volume_size" {
```
mattdean3-nhs: could we add a `powerbi_gw_` prefix to the var names? Just to make it clearer what they're used for.

mattdean3-nhs (approved):
LGTM

PR #1140, opened 2026-02-10
Feature/safo6 nrl 721 seed sandbox data

sandyforresternhs:
- Deletes all items in a DynamoDB table and reseeds the table with X (default = 2) pointers of each type for X (default is 2) custodians
- The scripts have been implemented to allow for execution either locally or via the lambda
- When running locally, the delete_all_table_items.py and seed_sandbox_table.py scripts can be run independently or via the orchestrator script reset_sandbox_table.py
- The lambda handles the orchestration of the delete and seed scripts (index.py), so the reset_sandbox_table.py script is not required for the lambda
- The lambda is deployed account-wide to dev and/or test (since we wouldn't want to reseed prod tables) and only when a table in the account has been specified for reseeding
- Logs showing the lambda working successfully can be seen [here](https://360016957465-rdeomcad.eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fnhsd-nrlf--dev--sandbox-seeder/log-events/2026$252F02$252F17$252F$255B$2524LATEST$255D79f615325d694728adfd7a13783facb7)
- Because it's account-wide we have needed to make the lamda layers available in the a/c wide infrastructure


scripts/tests/test_delete_all_table_items.py
```diff
@@ -0,0 +1,267 @@
...
+
+        mock_resource = MagicMock()
+        mock_resource.Table.return_value = mock_table
+        mock_boto3.resource.return_value = mock_resource
+
+        with pytest.raises(SystemExit):
+            delete_all_table_items("nonexistent-table")
+
+        mock_exit.assert_called_once_with(1)
+        mock_print.assert_called_with("Error: Table 'nonexistent-table' does not exist")
```
mattdean3-nhs: oooof - nice! 💪 am liking this mocking of exit and print.... definitely going to borrow that approach 😁

scripts/seed_sandbox_table.py
```diff
@@ -0,0 +1,393 @@
...
+import os
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+import boto3
+from botocore.exceptions import ClientError
+
+# Needed for when the script is run in Lambda where modules are in scripts subdirectory
```
mattdean3-nhs, line 18: I think these path/import issues could be fixed by adjusting the lambda runtime environment a little - but not a biggie for now, happy to leave as is and come back to fix later if ever needed. 👍

terraform/account-wide-infrastructure/dev/lambda__seed-sandbox.tf
```diff
@@ -0,0 +1,28 @@
+# Lambda to reset specified DynamoDB tables with seed test data on a schedule
+# Uses account-wide Lambda layers
+# Only deploys if tables are specified
+
+locals {
+  # take these out once development & testing is complete
+  seed_table_names = ["nhsd-nrlf--sandy-dev-pointers-table", "nhsd-nrlf--sandycopy-dev-pointers-table"]
```
mattdean3-nhs: probs remove sandycopy from list? or planning to do that after some testing?

terraform/account-wide-infrastructure/modules/seed_sandbox_lambda/eventbridge.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_cloudwatch_event_rule" "event_rule" {
+  name        = "${var.prefix}--event_rule"
+  description = "Rule to clear and reseed sandbox data"
+  #   Set this to weekly once development and testing is complete
+  #   schedule_expression = "cron(0 2 ? * SUN *)" # 2am UTC, every Sunday
+  schedule_expression = "cron(0/15 9-17 ? * MON-FRI *)" # Every 15 minutes, 9am-5pm UTC, Monday-Friday
```
mattdean3-nhs: worth making this schedule a module var? that way we could have different schedules for dev and test if we want. Not sure if it's worth it though, up to you.

terraform/account-wide-infrastructure/modules/seed_sandbox_lambda/iam.tf
```diff
@@ -0,0 +1,56 @@
...
+        ]
+        Resource = [for table_name in var.table_names : "arn:aws:dynamodb:${var.region}:*:table/${table_name}"]
+      },
+      {
+        Effect = "Allow"
+        Action = [
+          "kms:Decrypt",
+          "kms:DescribeKey"
+        ]
+        Resource = "*"
```
mattdean3-nhs: do we have an option to restrict this to a specific key? rather than all keys?

terraform/account-wide-infrastructure/test/lambda__seed-sandbox.tf
```diff
@@ -0,0 +1,28 @@
+# Lambda to reset specified DynamoDB tables with seed test data on a schedule
+# Deployed at account level to avoid duplication across workspaces
+# Uses account-wide Lambda layers
+# Only deploys if tables are specified
+
+locals {
+  seed_table_names = ["nhsd-nrlf--int-sandboxcopy-pointers-table"] # Change to int-sandbox-pointers-table when ready
```
mattdean3-nhs: change to correct table name before merge (or add the correct one and keep both for now)

mattdean3-nhs (commented):
This is EPIC 💪 nice work!

A few passing comments that I'd be happy to ignore.

Only one that we might need to address is the "*" wildcard used on the KMS key access for the new lambda - do we have an option to restrict that?

terraform/account-wide-infrastructure/dev/lambda__seed-sandbox.tf
sandyforresternhs: I left those two in for the time being just so that we could see it works with multiple tables, and we can see in the logs they're both getting reseeded.  I think the testers will probably want to add and remove tables, but I'm assuming that once we're all happy to merge we can remove the tables listed for dev completely, just leaving the int-sandbox table for test

terraform/account-wide-infrastructure/modules/seed_sandbox_lambda/eventbridge.tf
sandyforresternhs: Great call, yep will do that 👍

terraform/account-wide-infrastructure/test/lambda__seed-sandbox.tf
sandyforresternhs: Yep, I was thinking just before merge assuming we're happy for int to get cleared - not sure if we want to give customers the heads up first?

terraform/account-wide-infrastructure/modules/seed_sandbox_lambda/iam.tf
sandyforresternhs: Good catch, let's do that 👍

mattdean3-nhs (approved):

PR #1141, opened 2026-02-10
NRL-1938 Add CI policies to mgmt account

mattdean3-nhs:
All polices that have been added are unchanged from those defined in the NRLF-CI repo.

Other changes include:
- Create a ci-data bucket that can be used for workflow files and switch workflows to use it
- Add a var for the org-wide NHSDeveloper role name
- Lock-down the codebuild ECR access to only the read actions needed by codebuild (hopefully)

Once we're happy with this PR, I'll destroy the TF state from the NRLF-CI repo and then re-create the policies as part of the mgmt acc-wide deployment.


sandyforresternhs (dismissed):

katebobyn-nhs (approved):
reviewed verbally in ops catchup

PR #1146, opened 2026-02-13
 [NRL-1922] Get Sonarcloud reporting working as required

mattdean3-nhs:
- Update sonarcloud config to remove old config and bits that didn't work
- Added new "PR Checks" workflow that does build + test + sonarqube scan

There might be a bit more to do after this is merged as we need this on develop before we can do the audit of issues and resolve things.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -38,7 +38,7 @@ jobs:
           path: dist
 
   sonar:
-    name: SonarQube analysis and quality gate check
+    name: SonarQube scan
     runs-on: ubuntu-latest
     needs: build
     environment: pull-request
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,75 @@
+name: Run PR checks
```
sandyforresternhs, line 1: How about sonarqube checks for develop

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Get build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: build-artifacts
+          path: dist
+
+      - name: DEBUG - Check things are set up correctly
```
sandyforresternhs: Can this come out now

.github/workflows/pr-checks.yml
mattdean3-nhs: Done - have added it to the daily build workflow for develop

.github/workflows/pr-checks.yml
mattdean3-nhs: yep - reeeeemoved

sandyforresternhs (approved):

PR #1161, opened 2026-03-04
[NRL-1996] Remove perms lambda layer from acc-wide infra deploy

mattdean3-nhs:


.github/workflows/deploy-account-wide-infra.yml
sandyforresternhs, line 1: Should we bin off lines 105 & 145 as well?

mattdean3-nhs: Yep - have binned those too :heavy_check_mark:

sandyforresternhs (approved):

PR #1165, opened 2026-03-06
Feature/safo6 nrl 1949 assign pointer type perms via new model

sandyforresternhs:


scripts/get_s3_permissions.py
```diff
@@ -49,6 +49,44 @@ def add_test_files(folder, file_name, local_path):
...
+                "z00z-y11y-x22x",
+                "RX898",
+                [PointerTypes.MENTAL_HEALTH_PLAN.value],
+            ),  # http://snomed.info/sct|736253002
+        ],
+        "producer": [
+            (
+                "z00z-y11y-x22x",
+                "RX898",
+                [PointerTypes.MENTAL_HEALTH_PLAN.value],
```
katebobyn-nhs: Just thinking if all the test scenarios in the v2 feature will use these fixed perm sets, it might be helpful to set different type permissions for the producer vs consumer. That way the tests would prove that the consumer side of an application does not have any access to things permissioned only on the producer side (and vice versa) ?

tests/features/producer/v2-permissions-by-pointer-type.feature
```diff
@@ -0,0 +1,128 @@
...
+              "system": "https://fhir.nhs.uk/ValueSet/NRL-ResponseCode",
+              "code": "RESOURCE_DELETED",
+              "display": "Resource deleted"
+            }
+          ]
+        },
+        "diagnostics": "The requested DocumentReference has been deleted"
+      }
+      """
+    And the resource with id 'DK94-111-DeleteDocRefTest1' does not exist
```
katebobyn-nhs: This wasn't the ID of the one that was deleted here, think we should be asserting `RX898-111-DeleteDocRefTest1` does not exist

tests/features/producer/v2-permissions-by-pointer-type.feature
```diff
@@ -0,0 +1,128 @@
...
+      | property        | value                          |
+      | subject         | 9278693472                     |
+      | status          | current                        |
+      | type            | 736253002                      |
+      | category        | 734163000                      |
+      | custodian       | RX898                          |
+      | author          | HAR1                           |
+      | url             | https://example.org/my-doc.pdf |
+      | practiceSetting | 788002001                      |
+
```
katebobyn-nhs, line 86: Happy path looks good, could we have a negative 'create' scenario as well to prove types not listed are not permitted for Producers? (I'd expect a 403) 
This case would be useful to have as a baseline IMO as when we augment the permissions model with access controls and interactions, I expect the OperationOutcome diagnostics could change for some scenarios.

tests/features/producer/v2-permissions-by-pointer-type.feature
```diff
@@ -0,0 +1,128 @@
...
+      | id          | RX898-1111111111-SearchNHSDocRefTest1 |
+      | subject     | 9999999999                            |
+      | status      | current                               |
+      | type        | 736253002                             |
+      | category    | 734163000                             |
+      | contentType | application/pdf                       |
+      | url         | https://example.org/my-doc.pdf        |
+      | custodian   | RX898                                 |
+      | author      | X26                                   |
+    And the Bundle does not contain a DocumentReference with ID 'SG4-1111111111-SearchNHSDocRefTest3'
```
katebobyn-nhs, line 162: Final request (I think) - case where there is no permissions file for the org (within the app)? 
e.g. same App ID but some new org ABC0 which does not get populated in your setup script
So that the tests will prove that the type permissions are org-specific

scripts/get_s3_permissions.py
sandyforresternhs: Ooh yep, good call, will update...

tests/features/producer/v2-permissions-by-pointer-type.feature
sandyforresternhs: Good catch - sorted :)

tests/features/producer/v2-permissions-by-pointer-type.feature
sandyforresternhs: Good call, will do 👍

katebobyn-nhs (approved):
Super, happy for this to be merged ASAP so we have some regression tests to work with going forward :)

PR #1168, opened 2026-03-11
Feature/safo6 nrl 1986 enable allow all types

sandyforresternhs:
- PermissionsPolicy added to ConnectionMetadata to keep v2 permissions separate
- Endpoints updated to factor in the new PermissionsPolicy object
- V2 access control enabled for allow_all_types


tests/utilities/api_clients.py
mattdean3-nhs, line 238: Could update this with constants too to match ConsumerTestClient - no biggie though, functionally the same

layer/nrlf/core/decorators.py
```diff
@@ -159,11 +160,21 @@ def _load_v2_connection_metadata(headers: Dict[str, str], path: str):
     logger.log(LogReference.HANDLER004e)
     pointer_permissions = get_pointer_permissions_v2(metadata, path)
 
-    metadata.pointer_types = pointer_permissions.get("types", [])
+    metadata.nrl_permissions_policy = PermissionsPolicy.model_validate(
+        pointer_permissions
+    )
+
+    if "allow_all_types" in metadata.nrl_permissions_policy.access_controls:
```
mattdean3-nhs: Could use AccessControls.allow_all_types enum value instead of literal "allow_all_types" - again, no biggie, it does the same thing

mattdean3-nhs (dismissed):
Approved - all looks good to me. There's a couple super minor niggles but happy to leave.

tests/utilities/api_clients.py
sandyforresternhs: Good call, updated 👍

layer/nrlf/core/decorators.py
sandyforresternhs: Good call, updated 👍

mattdean3-nhs (approved):
🙌

PR #1195, opened 2026-03-31
NRL-2100 Update smoke tests to run for both v1 & v2 perms

anjalitrace2-nhs:
- create v2 and v1 permissions for smoke tests in get s3 perms script
- add v1 specific smoke tests - to be deleted once v1 phased out
- found & resolved a bug where the 1DSync test never triggered the internal mode runner and just ran as a normal producer
- unable to run public 1DSync tests realistically because we don't have a seperate 1DSync apigee app - no worse off for now. Public tests pass against dev (not run in pipeline) but as a normal producer doing upsert.

image


mattdean3-nhs (approved):
LGTM 👍

PR #1197, opened 2026-04-08
NRL-2099 manage permissions script (parity)

anjalitrace2-nhs:
- renamed existing manage permissions script -> manage_permissions_v1.py
- manage_permissions.py is for managing v2 permissions now
- for v2 permissions, you can use this script to:
    - list apps & orgs and show their permissions
    - list assignable pointer types & access controls (limited to currently supported access controls)
    - add and remove supported permissions - only types & access controls currently. Will not let you remove a permission not already assigned or add a permission already assigned.
    - clear permissions
- This is hopefully easily extensible for anyone adding support for new future permissions e.g. interactions, categories. Will just need to add to the list of supported keys `currently_supported_permission_keys` and fill in any `PERMISSION_KEY_ATTRIBUTES`

### DIFF
--- a/scripts/manage_permissions.py
+++ b/scripts/manage_permissions.py
@@ -28,9 +28,9 @@ COMPARE_AND_CONFIRM = (
 )
 
 currently_supported_access_controls = [
-    AccessControls.ALLOW_ALL_TYPES,
-    AccessControls.ALLOW_OVERRIDE_CREATION_DATETIME,
-    AccessControls.ALLOW_SUPERSEDE_WITH_DELETE_FAILURE,
+    AccessControls.ALLOW_ALL_TYPES.value,
+    AccessControls.ALLOW_OVERRIDE_CREATION_DATETIME.value,
+    AccessControls.ALLOW_SUPERSEDE_WITH_DELETE_FAILURE.value,
 ]
 
 
@@ -126,6 +126,7 @@ def list_apps(supplier_type: SupplierType) -> None:
     print(f"There are {len(app_level_perm_files)} apps with app-level permissions:")
     for app_level in app_level_perm_files:
         print(f"- {app_level}")
+    print()
 
 
 def list_orgs(supplier_type: SupplierType, app_id: str) -> None:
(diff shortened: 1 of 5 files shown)


scripts/manage_permissions.py
```diff
@@ -61,200 +97,488 @@ def _get_perms_from_s3(file_key: str) -> str | None:
...
     if not apps:
-        print("No applications found in the bucket.")
+        print(f"👀 No applications found in the {nrl_env} bucket.")
         return
 
-    print(f"There are {len(apps)} apps in {nrl_env} env:")
-    for app in apps:
-        print(f"- {app}")
+    def there_are_x_apps(app_count: int):
+        is_are = "is" if app_count is 1 else "are"
```
sandyforresternhs: These warnings are right - the 'is' is giving me a syntax error when there are multiple apps...
`is_are = "is" if app_count == 1 else "are"
        s = "" if app_count == 1 else "s"`

scripts/manage_permissions.py
```diff
@@ -61,200 +97,488 @@ def _get_perms_from_s3(file_key: str) -> str | None:
...
-                f"No changes needed for {app_id}/{org_ods}. Current permissions match the new ones."
-            )
-            return
+    current_perms = _load_or_setup_perms(lookup_path)
+    current_permission_items: list = current_perms.get(permission_key, [])
 
+    already_added_items = [
+        item for item in items_to_add if item in current_permission_items
+    ]
+    if already_added_items:
```
sandyforresternhs, line 418: Definitely not a deal breaker but I'm wondering if we'd still want it to add any remaining ones that we asked it to add - for example when adding types all - would we want it to print the types already set but still add the ones that aren't?

scripts/manage_permissions.py
anjalitrace2-nhs: nice, will do 👀

scripts/manage_permissions.py
anjalitrace2-nhs: I think this is a good shout - I'll do the equivalent for the remove too e.g.

can't remove these ones that don't exist, but it'll still remove the ones that do. Bit more declarative! 👍

sandyforresternhs (approved):

PR #1200, opened 2026-04-14
[NRL-1920] Use gitleaks for secret scanning

mattdean3-nhs:


.github/workflows/daily-build.yml
```diff
@@ -72,6 +72,29 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Gitleaks secret scan
+        uses: gitleaks/gitleaks-action@ff98106e4c7b2bc287b24eaf42907196329070c7 # v2.3.9
+        env:
+          GITLEAKS_VERSION: 8.30.1
+          GITLEAKS_CONFIG: .gitleaks.toml
+          GITLEAKS_LICENSE: ${{ secrets.NHSAPP_GITLEAKS_LICENSE }}
```
katebobyn-nhs, line 93: Can we still use this? I heard it was no longer available to us

mattdean3-nhs: Yes - we can still use it - we do have to get our own license for gitleaks which I've done and this is now set up in the repo.

katebobyn-nhs (approved):

